PR #52, opened 2020-09-03
Fixes

YoussefHenna:


YoussefHenna:
Please test on IOS because I don't have access to a mac. And also review the SEED_DATA because I am unsure if it written correctly.

package.json
```diff
@@ -161,6 +164,7 @@
     "dateformat": "^3.0.3",
     "expo-font": "^8.1.1",
     "expo-step-indicator": "0.1.2",
+    "react-native-safe-area-context": "~3.0.7",
```
peterpme, line 167: `expo install react-native-safe-area-context` installs 3.1.7

YoussefHenna: just reran the command in the example app and still got `~3.0.7`. Are you sure?

YoussefHenna:
There is a bug with react-native-slider and ScrollView on web. Whenever the slider is in deep within a scroll view, touch inputs are captured by the scroll view and not passed on to the slider. This is possibly due to how react-native-web is somewhat buggy when it comes to touch handling and scrolling. 

However, since I restructured the app to a drawer layout the example app works great on web and android with slider working on both, just thought you should be aware of the issue. I have tried to resolve it with workarounds but nothing seemed to work.

peterpme:
Hey @YoussefHenna that makes sense! I have an idea for you if you're willing to try it

react native gesture handler's scrollview has a slider inside of a scrollview located here 

https://snack.expo.io/@adamgrzybowski/react-native-gesture-handler-demo

If you scroll down and see the "combo" example you can run it on the web preview of snack. If it works we can use that under the hood. Lmk what you think. Thanks!

YoussefHenna:
The link you sent, they seem to be using the default Slider from react-native. I opted to use the community version instead since it is the recommended one to use. This does solve the scroll view issue, but it introduced another issue. It goes into an infinite setState loop and crashes (on web only). This happens because we have all 3 sliders configured to use the state value `value` in the `SliderExample` which leads to an infinite `onValueChanged` being called. Seems to work fine on android but keeps crashing on web, only works if state is to be managed internally by component without outside interference. 

I decided to leave it as is for now using `react-native-slider`  until I can figure something out, because I can't seem to think of a solution for now.

example/src/App.tsx
```diff
@@ -111,23 +134,94 @@ export default class App extends React.Component {
...
+  renderScreen = (key: string, Screen: any) => {
+    return () => (
+      <Example key={key} title={key}>
+        <Screen />
+      </Example>
+    );
+  };
+
   render() {
+    const Drawer = createDrawerNavigator();
```
peterpme, line 146: you want this to live above the component

peterpme:
Web: Datepicker crashes when clicking it 
image

peterpme:
Something funky happened to (3 Columns 1v1 aspectRatio) and the other components that use it:

this is the branch:
image

this is master:
image

example/src/App.tsx
```diff
@@ -1,5 +1,14 @@
+import "react-native-gesture-handler";
```
peterpme, line 1: whats this needed for?

example/src/App.tsx
```diff
@@ -111,23 +134,94 @@ export default class App extends React.Component {
...
       );
     } else {
       return <AppLoading />;
     }
   }
 }
+
+const exampleStyles = StyleSheet.create({
+  mainParent: {
+    width: screenWidth,
```
peterpme, line 183: do you need to include these? shouldn't this just work without it since a view's implicit width is 100%?

is there a difference between this and making the width 100%?

example/src/App.tsx
```diff
@@ -111,23 +134,94 @@ export default class App extends React.Component {
...
+    height: screenHeight,
```
peterpme, line 184: same comment here

src/components/RadioButtonFieldGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+    </View>
+  );
+};
+const styles = StyleSheet.create({
+  label: {
+    fontSize: 16,
+    color: "#000",
+  },
+});
+export default RadioButtonFieldGroup;
```
peterpme, line 89:
```suggestion
export default withTheme(RadioButtonFieldGroup);
```

need to wrap this in the theme

src/components/RadioButtonFieldGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+  key: string;
+  value: string;
+}
+
+interface Props {
+  options: RadioButtonOption[];
+  label?: string;
+  onSelect?: (value: string) => void;
+  value?: string;
+  radioButtonColor?: string;
```
peterpme, line 23: the colors should all be using the theme values, ie: "primary", "secondary", etc.

src/components/RadioButtonFieldGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+          </View>
+        );
+      })}
+    </View>
+  );
+};
+const styles = StyleSheet.create({
+  label: {
+    fontSize: 16,
+    color: "#000",
```
peterpme, line 86: we need these values to be driven by theme (both fontSize and color)

you can use theme.typography.headline4 for the label

src/components/RadioButtonFieldRow.tsx
```diff
@@ -0,0 +1,77 @@
...
+  onPress?: () => void;
+  selected: boolean;
+  theme: typeof themeT;
+}
+
+const RadioButtonFieldRow: React.FC<Props> = ({
+  label,
+  theme: { colors },
+  backgroundColor = colors.background,
+  labelColor = "#000",
```
peterpme, line 31: we need these values to be consumed by the theme and not hex

src/components/RadioButtonFieldRow.tsx
```diff
@@ -0,0 +1,77 @@
...
+  labelStyle?: StyleProp<TextStyle>;
+  onPress?: () => void;
+  selected: boolean;
+  theme: typeof themeT;
+}
+
+const RadioButtonFieldRow: React.FC<Props> = ({
+  label,
+  theme: { colors },
+  backgroundColor = colors.background,
```
peterpme, line 30: oh that's a neat trick! does this really work?

src/components/Slider.tsx
```diff
@@ -1,24 +1,26 @@
 import * as React from "react";
 // TODO replace this with community slider
```
peterpme, line 2: you can get rid of this TODO :smile:

src/components/RadioButtonFieldGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+  labelStyle,
+  itemStyle,
+  itemLabelStyle,
+  itemBackgroundColor,
+}) => {
+  let initValue = "";
+  if (value) {
+    initValue = value;
+  } else if (options.length > 0) {
+    initValue = options[0].value;
```
peterpme, line 51: i think its ok if there is no default value set (that should be possible)

src/components/Carousel.tsx
```diff
@@ -47,7 +50,7 @@ const Carousel: React.FC<Props> = ({
   resizeMode = "cover",
   dotColor,
   theme: { colors, spacing },
-  style = { height: 250 },
+  style = { height: screenWidth * 0.5 },
```
peterpme, line 53: is there a reason the default can't be a height of 250?

src/components/Carousel.tsx
```diff
@@ -75,7 +77,7 @@ const Carousel: React.FC<Props> = ({
         onScroll={handleScroll}
         horizontal
         pagingEnabled
-        showsHorizontalScrollIndicator={false}
+        showsHorizontalScrollIndicator={Platform.OS === "web"}
```
peterpme, line 80: good call

src/components/Stepper.tsx
```diff
@@ -11,13 +17,21 @@ interface Props {
...
+  borderRadius?: number;
+  typeStyle?: StyleProp<TextStyle>;
 }
 
 const Stepper: React.FC<Props> = ({
   value = 0,
   style,
   onChange,
   theme: { colors, typography, spacing },
+  iconSize = Config.stepperButtonSize,
```
peterpme, line 31:
```suggestion
  iconSize = 24,
```

the Config file is kinda worthless

src/components/Stepper.tsx
```diff
@@ -55,15 +75,16 @@ const Stepper: React.FC<Props> = ({
             color: colors.medium,
             marginHorizontal: spacing.medium,
```
peterpme, line 76:
```suggestion
            marginHorizontal: spacing.small,
```

src/components/Stepper.tsx
```diff
@@ -84,6 +105,36 @@ export const SEED_DATA = [
...
+        group: GROUPS.basic,
+        label: "Current Icon Color",
+        description: "The color of the icons",
+        editable: true,
+        required: false,
+        formType: FORM_TYPES.color,
+        propType: PROP_TYPES.THEME,
+        defaultValue: null,
+      },
+      borderRadius: {
```
peterpme, line 128:
```
...BORDER_RADIUS_MODE
```

from component-types.ts

src/components/Stepper.tsx
```diff
@@ -84,6 +105,36 @@ export const SEED_DATA = [
         ...FIELD_NAME,
         defaultValue: "stepperValue",
       },
+      iconSize: {
```
peterpme, line 108: really close! a few extra props:
min: 0, max: 60, step: 1

src/components/Stepper.tsx
```diff
@@ -84,6 +105,36 @@ export const SEED_DATA = [
...
+      },
+      iconColor: {
+        group: GROUPS.basic,
+        label: "Current Icon Color",
+        description: "The color of the icons",
+        editable: true,
+        required: false,
+        formType: FORM_TYPES.color,
+        propType: PROP_TYPES.THEME,
+        defaultValue: null,
```
peterpme, line 126:
```suggestion
        defaultValue: "strong",
```

src/components/Stepper.tsx
```diff
@@ -84,6 +105,36 @@ export const SEED_DATA = [
...
       },
+      iconSize: {
+        group: GROUPS.basic,
+        label: "Current Icon Size",
+        description: "The size of the icons",
+        editable: true,
+        required: false,
+        formType: FORM_TYPES.number,
+        propType: PROP_TYPES.NUMBER,
+        defaultValue: null,
```
peterpme, line 116:
```suggestion
        defaultValue: 24,
```

src/components/Stepper.tsx
```diff
@@ -11,13 +17,21 @@ interface Props {
...
 }
 
 const Stepper: React.FC<Props> = ({
   value = 0,
   style,
   onChange,
   theme: { colors, typography, spacing },
+  iconSize = Config.stepperButtonSize,
+  iconColor = colors.strong,
+  borderRadius = 0,
```
peterpme, line 33:
```suggestion
  borderRadius = borderRadius.button
```

comes from theme

src/components/Stepper.tsx
```diff
@@ -11,13 +17,21 @@ interface Props {
...
+  iconColor?: string;
+  borderRadius?: number;
+  typeStyle?: StyleProp<TextStyle>;
 }
 
 const Stepper: React.FC<Props> = ({
   value = 0,
   style,
   onChange,
   theme: { colors, typography, spacing },
```
peterpme, line 30:
```suggestion
  theme: { colors, typography, spacing, borderRadius },
```

peterpme:
image

the toolbar / menu icon seem to grow with the height of the device, can these be static? call it 50px for the toolbar and then maybe 32px for the menu icon? whatever looks good as long as it doesn't grow :smile:

package.json
peterpme: You're right, thanks!

example/src/App.tsx
YoussefHenna: React navigation. Required to be the first import. Removing might seem fine, but they warn in might break in release versions.

example/src/App.tsx
YoussefHenna: From my testing, web misbehaved when using percentages (inconsistent with mobile), so I opted to use this approach.

src/components/RadioButtonFieldRow.tsx
YoussefHenna: yes

example/src/App.tsx
YoussefHenna: Scratch that, works fine everywhere except for images. Images do not appear when set using percentages (on web I mean). Can remove it everywhere except with images.

src/components/Carousel.tsx
YoussefHenna: When it was hard-coded to 250, it looked weird on large width devices (tablets or pc for web), remember the issue with aspect ratio. This way it will look more or less the same aspect ratio on all devices.

example/src/App.tsx
YoussefHenna: Managed to remove them completely. Ignore all that

YoussefHenna:
Managed to address all the issues. Let me know if I missed anything.

example/src/App.tsx
```diff
@@ -111,23 +134,94 @@ export default class App extends React.Component {
...
+                    activeTintColor: "rgba(90, 69, 255, 1)",
+                  }}
+                  initialRouteName="Avatar"
+                >
+                  {Object.entries(ROUTES).map(([key, Screen]) => {
+                    return (
+                      <Drawer.Screen
+                        key={key}
+                        name={key}
+                        component={this.renderScreen(key, Screen)}
```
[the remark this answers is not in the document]
adnelson: That + remove the `renderScreen` function

adnelson (approved):
small suggestion, looks good

example/src/App.tsx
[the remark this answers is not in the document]
adnelson: That + remove `renderScreen`

adnelson (commented):
small suggestion, looks good

adnelson:
Oops, sorry for the double comment.

example/src/App.tsx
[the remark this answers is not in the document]
YoussefHenna: key here is not meant as the `key` used by components, it's just a name used to indicate the `(key,value)` pair of the object. It's just a string passed to be use as the title.

PR #62, opened 2020-09-16
Progress indicator theme fix

YoussefHenna:
I know you asked for the current item to be same bg as finished, but the library used doesn't seem to allow/support that. Best I could do was add a stroke on it to identify the that it is the current. This looks better in my opinion, but let me know if that is not what you want.

### DIFF
--- a/example/src/ProgressIndicatorExample.js
+++ b/example/src/ProgressIndicatorExample.js
@@ -12,7 +12,7 @@ function ProgressIndicatorExample({ theme }) {
           currentStepStrokeWidth={3}
           stepIndicatorSize={28}
           stepIndicatorLabelFontSize={12}
-          stepNumberFinishedColor={"strongInverse"}
+          stepNumberFinishedColor={"primary"}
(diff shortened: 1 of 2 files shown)


example/src/ProgressIndicatorExample.js
```diff
@@ -9,13 +9,13 @@ function ProgressIndicatorExample({ theme }) {
...
-          currentStepStrokeWidth={0}
+          currentStepStrokeWidth={3}
           stepIndicatorSize={28}
-          stepIndicatorLabelFontSize={16}
-          stepNumberFinishedColor={"#5a45ff"}
-          stepNumberUnfinishedColor={"#ffffff"}
-          unfinishedColor={"#eee"}
-          finishedColor={"#5a45ff"}
+          stepIndicatorLabelFontSize={12}
+          stepNumberFinishedColor={"strongInverse"}
```
peterpme, line 15:
```suggestion
          stepNumberFinishedColor={"primary"}
```

peterpme (approved):

PR #63, opened 2020-09-16
Added datepicker support for web

YoussefHenna:
used https://material-ui.com/components/pickers/#material-ui-pickers for web support

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "pod-install": "^0.1.7",
     "prettier": "^2.0.5",
     "react": "~16.9.0",
+    "react-dom": "^16.13.1",
     "react-native": "~0.62.2",
     "react-native-svg": "^12.1.0",
     "release-it": "^13.6.3",
@@ -167,7 +168,6 @@
     "dateformat": "^3.0.3",
     "expo-font": "^8.1.1",
     "expo-step-indicator": "0.1.2",
-    "react-dom": "^16.13.1",
     "react-native-safe-area-context": "~3.0.7",
     "react-native-slider": "^0.11.0",
     "react-native-typography": "^1.4.1"
--- a/package.json
+++ b/package.json
@@ -158,11 +158,16 @@
     ]
   },
   "dependencies": {
+    "@date-io/date-fns": "1.x",
     "@draftbit/react-theme-provider": "^2.1.1",
+    "@material-ui/core": "^4.11.0",
+    "@material-ui/pickers": "^3.2.10",
     "color": "^3.1.2",
+    "date-fns": "^2.16.1",
     "dateformat": "^3.0.3",
     "expo-font": "^8.1.1",
     "expo-step-indicator": "0.1.2",
+    "react-dom": "^16.13.1",
     "react-native-safe-area-context": "~3.0.7",
     "react-native-slider": "^0.11.0",
     "react-native-typography": "^1.4.1"
--- a/src/components/DatePicker.tsx
+++ b/src/components/DatePicker.tsx
@@ -2,7 +2,6 @@ import * as React from "react";
 import { View, StyleSheet, StyleProp, ViewStyle } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 
-import DateTimePicker from "@react-native-community/datetimepicker";
 import dateFormat from "dateformat";
 import { withTheme } from "../core/theming";
 
@@ -11,6 +10,8 @@ import Button from "./Button";
 import TextField, { Props as TextFieldProps } from "./TextField";
 import Touchable from "./Touchable";
 import theme from "../styles/DefaultTheme";
+//@ts-ignore
+import DateTimePicker from "./DatePickerComponent";
 
 interface Props extends TextFieldProps {
   style?: StyleProp<ViewStyle>;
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -158,11 +158,16 @@
...
+    "@date-io/date-fns": "1.x",
     "@draftbit/react-theme-provider": "^2.1.1",
+    "@material-ui/core": "^4.11.0",
+    "@material-ui/pickers": "^3.2.10",
     "color": "^3.1.2",
+    "date-fns": "^2.16.1",
     "dateformat": "^3.0.3",
     "expo-font": "^8.1.1",
     "expo-step-indicator": "0.1.2",
+    "react-dom": "^16.13.1",
```
peterpme, line 170: i think this has to live under `devDependencies`, let me get back to you in a bit

src/components/DatePicker.tsx
```diff
@@ -11,6 +10,8 @@ import Button from "./Button";
 import TextField, { Props as TextFieldProps } from "./TextField";
 import Touchable from "./Touchable";
 import theme from "../styles/DefaultTheme";
+//@ts-ignore
```
peterpme, line 13: any reason you need this here? is it because of the .native and .web extension? i think this will solve your issue:

https://stackoverflow.com/questions/44001050/platform-specific-import-component-in-react-native-with-typescript

src/components/DatePickerComponent.native.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import DateTimePicker from "@react-native-community/datetimepicker";
+
+interface Props {
+  value: Date;
+  onChange: (e: any, data?: any) => void;
+  mode: "date" | "time" | "datetime";
+  toggleVisibility: () => void;
+}
+const DatePickerComponent: React.FC<Props> = ({
```
peterpme, line 10: this'll need to include the withTheme stuff and theme variables so it matches accordingly

src/components/DatePickerComponent.web.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+interface Props {
+  value: Date;
+  onChange: (e: any, data?: any) => void;
+  mode: "date" | "time" | "datetime";
+  toggleVisibility: () => void;
+  isVisible: boolean;
+}
+
+const DatePickerComponent: React.FC<Props> = ({
```
peterpme, line 18: needs to use withTheme, etc

src/components/DatePicker.tsx
```diff
@@ -11,6 +10,8 @@ import Button from "./Button";
...
+import DateTimePicker from "./DatePickerComponent";
```
peterpme, line 14: you can call this DateTimePicker.native.tsx and DateTimePicker.web.tsx

src/components/DatePickerComponent.web.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import DateFnsUtils from "@date-io/date-fns";
+import {
+  MuiPickersUtilsProvider,
+  DatePicker,
+  TimePicker,
+  DateTimePicker,
+} from "@material-ui/pickers";
+
+interface Props {
```
peterpme, line 10: are the props between web and native the same? can we share those inside the main DatePicker file?

src/components/DatePickerComponent.native.tsx
YoussefHenna: this is just a wrapper for the component from the library (just to split web and native), it isn't style-able using any colors or styles. The actual DatePicker component is still used and wrapped in a withTheme. Wrapping this withTheme would be of no use I guess.

src/components/DatePicker.tsx
YoussefHenna: renamed to DateTimePicker.tsx and DateTimePicker.web.tsx. No ts errors, native picks up normal one and web picks up web

src/components/DatePickerComponent.web.tsx
YoussefHenna: that would lead to both DatePicker and DatePickerComponent to be importing each other which causes issues. I'll put it in a separate file and put them all in a folder, feel that's most appropriate.

src/components/DatePicker/DatePickerComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+      mode={mode}
+      onChange={(_event: any, data: any) => {
+        toggleVisibility();
+        onChange(null, data);
+      }}
+    />
+  );
+};
+
+export default DatePickerComponent;
```
peterpme, line 23: will the right colors show up given the theme? i think thats the last thing that's missing

YoussefHenna: The date picker I added ([material-ui](https://material-ui-pickers.dev/api/DatePicker)) and the old one ([react-native-community/datetimepicker](https://github.com/react-native-community/datetimepicker)) do not have any color customization or styling. So theme will have no affect on them, but that is not a bug I introduced or anything it was how it was and I don't think there is anything we can do about it.

peterpme:
:100: !! 

one small thign: on Web the old modal still appears slightly

(see the close button in the bottom right)
image

peterpme:
This bottom one (red box) doesn't work 
image

src/components/DatePicker/DatePickerComponent.tsx
peterpme: this will also need the theme for android (not ios though)

peterpme:
image

android just needs theme and we're good!

YoussefHenna:
Color cannot be changed dynamically on android. Check https://github.com/react-native-community/datetimepicker/issues/20

Can only be changed by hardcoding colors into android xml files. Library does not support it

YoussefHenna:
Red one works fine for me. Can you check again

YoussefHenna:
I made it so close button appears only on ios. Android & web have their own dedicated close button

PR #68, opened 2020-09-27
Issue 65 Fix

YoussefHenna:


src/components/AspectRatio.web.tsx
```diff
@@ -1,50 +1,40 @@
 import React from "react";
 import {
   View,
-  createElement,
   StyleSheet,
-  LayoutChangeEvent,
+  LayoutRectangle,
```
peterpme, line 5: what does this do?

src/components/AspectRatio.web.tsx
```diff
@@ -1,50 +1,40 @@
...
 interface Props {
   children?: React.ReactNode;
-  onLayout?: (event: LayoutChangeEvent) => void;
-  ratio: number;
   style?: StyleProp<ViewStyle>;
 }
 
-const AspectRatio: React.FC<Props> = ({ children, onLayout, ratio, style }) => {
-  const percentage = 100 / ratio;
+const AspectRatio: React.FC<Props> = (props) => {
```
peterpme, line 15: this component came from [react-native-web](https://github.com/necolas/react-native-web/issues/427#issuecomment-352474031)

do you know what differences your changes will make?

src/components/AspectRatio.web.tsx
```diff
@@ -1,50 +1,40 @@
...
-  LayoutChangeEvent,
+  LayoutRectangle,
   StyleProp,
   ViewStyle,
 } from "react-native";
 
 interface Props {
   children?: React.ReactNode;
-  onLayout?: (event: LayoutChangeEvent) => void;
-  ratio: number;
```
peterpme, line 14: i think we need ratio here still, no?

src/components/AspectRatio.web.tsx
YoussefHenna: Just an object with width and height

src/components/AspectRatio.web.tsx
YoussefHenna: It didn't seem to work at all, aspect ratio refused to work. Don't worry though I didn't make this up from my mind, I got it from [expo's website](https://baconbrix.gitbook.io/react-native-web/styling/aspect-ratio) just added some types to it to be valid typescript. This one works as intended. Maybe the old approach is outdated in newer expo versions.

src/components/AspectRatio.web.tsx
YoussefHenna: It uses ratio directly provided in the style, no need for additional prop

PR #72, opened 2020-10-16
Fixed aspect ratio for native devices

YoussefHenna:
For some reason, aspect ratio refused to work whenever a local image was used but worked fine with any network image. Wrapping it in a View and having aspect ratio set there seemed to do the trick, so that's what I changed. Whenever it is a local image and uses aspect ratio, it will automatically be wrapped in a View.

### DIFF
--- a/src/components/AspectRatio.web.tsx
+++ b/src/components/AspectRatio.web.tsx
@@ -21,9 +21,9 @@ const AspectRatio: React.FC<Props> = (props) => {
   if (layout) {
     const { width = 0, height = 0 } = layout;
     if (width === 0) {
-      style.push({ width: height * aspectRatio, height });
+      style.push({ width: height * (1 / aspectRatio), height });
     } else {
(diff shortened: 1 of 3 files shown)


peterpme:
Close! 

native device:
image


web (same width):
image


It looks like the aspectRatios are flipped on web vs. native. For example:

2/3 on web looks more correct:
image

but 2/3 on native looks off
image



3/2 on native looks correct:
image


but 3/2 on web looks off:
image

YoussefHenna:
Just spent an hour trying to figure out why they were reversed. Turns out all I need to do was `1/aspectRatio` instead of `aspectRatio`.

Anyways should be working as expected now.

YoussefHenna:
Also added Stylsheet.flatten in case someone were to pass style as an array rather than an object.

peterpme (approved):
LGTM, thanks!

PR #74, opened 2020-10-20
Added webview

YoussefHenna:
added webview using both https://github.com/react-native-webview/react-native-webview & https://github.com/react-native-web-community/react-native-web-webview

Webpack needs to be configured for it to work though, see the changes in the example app.

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -46,6 +46,8 @@ import StepperExample from "./StepperExample";
 
 import TextFieldExample from "./TextFieldExample";
 
+import WebViewExample from "./WebViewExample";
+
 import ProgressIndicatorExample from "./ProgressIndicatorExample.js";
 import {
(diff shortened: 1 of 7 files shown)


peterpme:
Was this working for you on the web? I get this error: 
image

peterpme:
I was taking a look at Expo AV and how they handle createElement, you might have to do something like this:

```
import createElement from 'react-native-web/dist/exports/createElement';
```

https://github.com/expo/expo/blob/master/packages/expo-av/src/ExponentVideo.web.tsx

YoussefHenna:
It was working for me. Will take a second look, maybe I missed something

YoussefHenna:
It works for me on web and android. Here is a screenshot from web. The same error your getting I got when webpack wasn't configured correctly, so make sure it matches and try again.
Screenshot_1

PR #84, opened 2020-10-29
Tests for components

YoussefHenna:


YoussefHenna:
Still a work in progress. Created so you can track my progress

YoussefHenna:
While creating the tests, some of them kept failing because of an issue in Icon.tsx. It has something to do with the `VectorIcons` imported from `@expo/vector-icons`. I tried taking a look to possibly fix it, but to be honest, I got lost in this code snippet.

`let VectorIcons: any;

try {
  // Optionally require vector-icons
  VectorIcons = require("@expo/vector-icons");
} catch (e) {
  if (
    // @ts-ignore
    global.__expo &&
    // @ts-ignore
    global.__expo.Icon
  ) {
    // Snack doesn't properly bundle vector icons from subpath
    // Use icons from the __expo global if available
    // @ts-ignore
    VectorIcons = global.__expo.Icon;
  } else {
    let isErrorLogged = false;

    // Fallback component for icons
    // @ts-ignore
    VectorIcons = ({ name, color, size, style, ...rest }) => {
      if (!isErrorLogged) {
        if (
          !/(Cannot find module|Module not found|Cannot resolve module)/.test(
            e.message
          )
        ) {
          console.error(e);
        }

        console.warn(
          `Tried to use the icon '${name}' in a component from '@draftbit/ui', but '@expo/vector-icons' could not be loaded.`,
          `To remove this warning, try installing '@expo/vector-icons'`
        );

        isErrorLogged = true;
      }

      return (
        <Text
          {...rest}
          style={[{ color, fontSize: size }, style]}
          // @ts-ignore
          pointerEvents="none"
        >
          □
        </Text>
      );
    };
  }
}`

`const IconSet = VectorIcons[iconSet];`

I don't really understand how this is supposed to work, so didn't want to mess with it in order to not mess it up even more. But the tests are failing because at the end `VectorIcons[key]` is not a component but rather undefined. Not sure why this happens, thought you might know. If not, then can you at least tell me what the above snippet is supposed to be doing? That could help.

If you are as clueless as me though, let me know and I'll dig into it until I figure something out.

YoussefHenna:
I've added tests for all the files

PR #86, opened 2020-12-01
fix(textfield): fix overlap of label and placeholder

babzcraig:
Fixed the issue where label text overlaps the placeholder for the TextField component.

Before:
Screen Shot 2020-11-30 at 9 01 34 PM

After:
Screen Shot 2020-11-30 at 9 01 46 PM


dhruska (approved):
Looks good!

PR #93, opened 2020-12-14
Fix DateTimePicker display in iOS 14

StartupSam:
Two small fixes in this PR. 

First, the current picker closes on every value change, which is a very bad UX when paired with a spinner. Fixed by only toggling visibility on change if the platform isn't iOS.

Second, in iOS 14, this component defaults to an inline datetime picker, which looks horrible in the space designed for a spinner. Fixed by defaulting the picker to a spinner on iOS.

I've already implemented this change in my local copy of Draftbit, and everything seems to work as it should.


StartupSam:
Note: This doesn't actually prevent the iOS datetime picker being dismissed problem. I thought it did, but seems I was mistaken.

peterpme:
Thanks @StartupSam !

PR #94, opened 2020-12-14
Fix Icon component on web

dhruska:
https://snack.expo.io/@donald/jigsaw-icon-error


publish.sh
```diff
@@ -1,19 +1,13 @@
 #!/usr/bin/env sh
 
+# normal package for react native
```
dhruska, line 3: Easier to publish UI package first rather than publish web and transform package.json back to UI version

scripts/prepare-web-pak.js
```diff
@@ -0,0 +1,26 @@
...
+      ...pak,
+      name: "@draftbit/" + name,
+    },
+    null,
+    2
+  );
+}
+
+function removeDependency(package, dependencyName) {
+  return package.replace(new RegExp(` +"${dependencyName}"[^\n]+\n`), "");
```
peterpme, line 16: this is technically json so you could remove the key instead of using regex

peterpme (approved):
LGTM

scripts/prepare-web-pak.js
dhruska: good call. fixed in 4511140474b858c381ce586bd657c2842d63b0fd

peterpme (approved):

PR #101, opened 2021-01-07
refactor: Bring expo-yarn-workspaces and add package folder

davesnx:
Apologies for the big PR, this PR removes the unimodules and brings expo-yarn-workspaces forcing us to bring a little yarn-workspaces structure into this repository. GitHub in this case fails miserably at showing the diff and some files that there's diff appears as just moving files, so, If you are interested in reviewing them, it's based on this diff: https://github.com/brentvatne/expo-av-example/commit/e3d4342ddcc5cd9ffc3fcebfbbff7f5ca173b04d

Aside from the big number of changes, there's duplicity on packagejson names, repo names and folder names, which we will cleanup once we ensure this changes work great with their clients.

- Adds a top level package.json with expo-yarn-workspaces
- Brings `example` and `react-native-jigsaw` under **packages/** where example points directly to "packages/react-native-jigsaw" under it's name "@draftbit/ui" see here: packages/example/src/App.tsx

cc @peterpme


peterpme (approved):
:100:

PR #105, opened 2021-01-12
Fix iOS DateTimePicker closing too early bug

StartupSam:
Fixes iOS DateTimePicker bug where the time selection modal would close every time the value changed (whenever the scroll wheel stopped moving). 1-line PR.


peterpme (approved):
Thank you @StartupSam !!! You are awesome!!!

PR #108, opened 2021-01-18
layout components

peterpme:


packages/react-native-jigsaw/src/components/Layout.tsx
```diff
@@ -0,0 +1,324 @@
...
+import {
+  GROUPS,
+  FORM_TYPES,
+  PROP_TYPES,
+  COMPONENT_TYPES,
+} from "../core/component-types";
+
+export function Center({
+  width,
+  height,
```
dhruska, line 15: Why pass `width` and `height` as props but also a style object? Should they just live in `style`?

peterpme: trying to make the api as small as possible, wdyt?

dhruska: works 4 me

dhruska (approved):
Excited to see this!!

davesnx (approved):

PR #109, opened 2021-01-21
@adnelson/email login screen

adnelson:
* EmailLoginScreen component
* DataContext for passing in callbacks for data operations


hooks/pre-commit
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
pckilgore, line 1: aye, we're close to reinventing lerna, might as well just use it sooner rather than later

packages/react-native-jigsaw/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@draftbit/ui",
+  "version": "41.0.0",
```
pckilgore, line 3: We keep the major version in sync with Expo, so unless I missed SDK 41, you're probably looking at `40.3.0` here.

packages/react-native-jigsaw/package.json
```diff
@@ -25,7 +26,11 @@
     "example": "yarn --cwd example",
     "pods": "cd example && pod-install --quiet",
     "bootstrap": "yarn example && yarn && yarn pods",
-    "postinstall": "expo-yarn-workspaces check-workspace-dependencies"
+    "postinstall": "expo-yarn-workspaces check-workspace-dependencies",
+    "set-up-hooks": "mkdir -p ../.git/hooks && cp -v ../../hooks/* ../../.git/hooks",
```
pckilgore, line 30: can we just use husky here?  Easier maintenance burden + very common in JS open source ecosystem so we can get some help with that.  Also, windows support.

packages/react-native-jigsaw/src/core/DataContext.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const defaultDataContext: DataContextType = {
+  signInWithEmailAndPassword: () => {
+    throw new Error("Signin has not been implemented for this app");
+  },
+  signOut: () => {
+    throw new Error("Signout has not been implemented for this app");
+  },
+};
+
+export const DataContext = React.createContext(defaultDataContext);
```
pckilgore, line 20: I think something like this is a really good idea, I only want to ask how tightly you think your changes might be tied to this exact API, because this is something I think we should align on as an engineering team.

If it's isolated to firebase stuff and you're pretty confident it can be unwound and are just experiment then :rocket:

pckilgore (commented):
The version number thing is probably blocking, some tooling suggestions because this is a little different scenario than our private product repo, and questions above.

packages/react-native-jigsaw/package.json
adnelson: Oh good call actually this is semantically a minor version change anyway come to think of it. 👍

packages/react-native-jigsaw/package.json
adnelson: maybe? but considering it's a monorepo and husky wasn't having fun with that before, I figured we could just follow the same approach

packages/react-native-jigsaw/src/core/DataContext.tsx
adnelson: Totally, this is all subject to change going forward. For the time being it's pretty much just exploratory

packages/react-native-jigsaw/package.json
pckilgore: Yeah, that's because we were running multiple different pre-commit scripts, here's there's really just one.  Its easy enough to do later.

pckilgore (approved):
a-ok with me

PR #112, opened 2021-01-22
Split hidden package web json into its own module

davesnx:
## Todos

- [x] Remove publish.sh
- [x] Create a web module which:
  - [x] Removes Icon and AudioPlayer
  - [x] Keeps expo-av and expo icons out of their package.json. Now they are both separated.
- [x] Test collector script
- [x] Run a mock and test the components endpoint
- [x] Automate publication
- [x] Remove release-it

## Summary

This PR introduces a few changes that are substantial to the project and I would love everyone's approval to move it forward and any feedback required to have this into a better status.

In master, in order to publish `@draftbit/ui` and `@draftbit/web` we use the same package.json with a few mutations (Remove some web-only files and nullify some other web-only components) and the publish.sh script would do the mutations and publish both.

In order to fix this, I created a `/web` under packages where a package.json lives with those modifications set in stone under a new package.json and publish it separately from /ui.

> In case you were wondering, everytime that we mention @draftbit/ui we mean `packages/react-native-jigsaw` (the folder), not the repo. So, In a future PR I will align and rename `packages/react-native-jigsaw` to `ui`.

This PR introduces:

- Publication for each web/ui module on master (following the version field under it's package.json)
- Publication for each web/ui module on PR, each PR will release a release candidate version to the registry where it looks like: `40.3.0-rc89b4d84.0` and flag it as `next`. This PR already published that version in `https://www.npmjs.com/package/@draftbit/ui`.
- Runs the collector script on CI and have a dummy contract testing where the CI runs a mock-server and ensures that some request happens against `localhost:3001/components`.

The introduction of 2 packages.json makes things a little tedious to remember and to keep sync, so I added a comment under `packages/web/package.json`: `"NOTE": "This package.json should be in sync with @draftbit/ui's package.json"`.


packages/react-native-jigsaw/scripts/parser.js
```diff
@@ -1,10 +1,8 @@
-const fs = require("fs");
+const { promises: fs } = require("fs");
```
peterpme, line 1: seems like a node v15 & above thing? is that true?

if so then we def need to bump our nodejs version in the CI

didn't know about this, is awesome!

packages/react-native-jigsaw/db.json
```diff
@@ -0,0 +1,3 @@
+{
+  "components": []
```
peterpme, line 2: whats this?

packages/web/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as JigsawUI from "../react-native-jigsaw/src/index";
```
peterpme, line 1: with workspaces you shouldn't need to do this, right? you should be able to do `import * as JigsawUI from '@draftbit/ui'` right?

packages/web/package.json
```diff
@@ -0,0 +1,157 @@
+{
```
peterpme, line 1: im not sure if `web` is actually the right name for this because its not for the web, its for the builder only. I think a better name is `builder` to be specific for anyone else

peterpme:
if not in this exact PR, then immediately afterwards lets rename the react-native-jigsaw to `ui`, the sooner we do that, the better

packages/react-native-jigsaw/scripts/parser.js
davesnx: `require("fs/promises")` is from Node v14, but the namespace promises have been since 12 I believe.

packages/web/package.json
davesnx: Just name it web since the package published is web, I don't have any particular preference rather than keep consistency with the names published on NPM/Github repo.

packages/web/index.tsx
davesnx: I could do `"@draftbit/ui"` but then I would rely on whatever is under the "main" package.json for UI, so I explicitly make that dependency

Happy to annotate that or change it to rely on "main".

packages/react-native-jigsaw/db.json
davesnx: This is a mock server that creates an endpoint /components mimicking what we have on the backend in order to test the collector script.

packages/react-native-jigsaw/package.json
```diff
@@ -55,6 +48,7 @@
     "dateformat": "^3.0.3",
     "expo-av": "^8.7.0",
     "expo-yarn-workspaces": "^1.3.0",
+    "json-server": "^0.16.3",
```
pckilgore, line 51: devDependency

packages/web/package.json
pckilgore: I agree that even if its the wrong name we already chose it a while ago and the costs of changing it now >>>> costs of having a slightly inaccurate name.

packages/web/package.json
```diff
@@ -0,0 +1,157 @@
...
+  "NOTE": "This package.json should be in sync with @draftbit/ui's package.json",
```
pckilgore, line 2: I think having to manually sync these is going to be almost as bad as a mutat-ey script.

Any way to avoid that or reduce that? For example, have this depend on @draftbit/ui or a @draftbit/core package?

packages/web/tsconfig.json
```diff
@@ -0,0 +1,36 @@
+{
```
pckilgore, line 1: Any way to share common config options like babel (where you can just have a core config, grab it, spread it in and change what you need)? 

Write our own extends?

pckilgore (commented):
I'll be honest--as much as I thought this might simplify stuff I regret suggesting it. 

I think having to keep all this stuff in sync manually and know all this context is probably the same difficulty in maintaining jigsaw as the current script-based solution.

I'm open to more ways to share configs where we can vs. repeat them, but I sort of lean against doing this at this point.

davesnx:
Totally, having both sync could be painful. Would be worth it explore a package.json generation script for both. Aside from the repetition and sync there, is there any other concern? @pckilgore

packages/web/package.json
davesnx: The only way I thought, if we want to keep the dependency graph is, to generate both package.jsons from a template. This would solve the repetition but would add another layer of complex.

davesnx: Another approach, but this would require more changes, could be to keep web/ui part of each component and have a change on the build tool where are both exposed under the same package.

packages/web/tsconfig.json
davesnx: For sure, let me try to fix this.

packages/web/index.tsx
```diff
@@ -1,3 +1,5 @@
+/* This points directly to src/index instead of "@draftbit/ui" because we don't want to rely
+on the package.json "main" */
```
pckilgore, line 2: Would be helpful to explain _why_ we don't want to rely on the package.json main?

packages/react-native-jigsaw/tsconfig.json
```diff
@@ -1,37 +1,3 @@
...
-    "strictFunctionTypes": true,
-    "alwaysStrict": true,
-    "noUnusedParameters": true,
-    "noImplicitAny": true,
-    "resolveJsonModule": true,
-    "skipLibCheck": true,
-    "strict": true,
-    "target": "esnext"
-  }
+  "extends": "../../tsconfig.json"
```
pckilgore, line 2: nice

packages/web/package.json
```diff
@@ -0,0 +1,157 @@
...
+  },
+  "@react-native-community/bob": {
+    "source": "src",
+    "output": "lib",
+    "targets": [
+      "commonjs",
+      "module",
+      "typescript"
+    ]
+  }
```
pckilgore, line 156: If we're going to manually support this... do we actually need any of this?

packages/web/index.tsx
davesnx: Oh, good point thought was obvious. Updated!

pckilgore:
Maybe document new publication process -- even if that it "just bump the version and merge to master and /web and /ui will be published automatically to npm by CI"?

Otherwise OK with me

packages/web/package.json
davesnx: Good point, everything except `"@react-native-community/bob"` I believe we can remove it

PR #116, opened 2021-01-29
Lerna

pckilgore:
Combine @davesnx  and @adnelson 's improvements with straightforward DX using lerna.

Most of the changes are moved files.

I will take responsibility for supporting these changes going forward.

DX now is simple, see `quickstart` in Readme.

In general, the idea here is to support actual dependency injection rather than faking it with publish scripts, but still avoid duplicating lots of code.

**I highly suggest reading the README files first when reviewing, they explain the lay of the land**


CONTRIBUTING.md
```diff
@@ -55,9 +58,45 @@ Remember to add tests for your change if possible. Run the unit tests by:
...
+editable: true,
+required: false,
+formType: FORM_TYPES.flatArray,
+propType: PROP_TYPES.STRING,
+defaultValue: "highQuality",
+```
+
+`src/core/component-types.js` will show you the different GROUPS, PROP_TYPES,
+FORM_TYPES, and other fields you might need.
+
```
pckilgore, line 95: Moved from base package readme

CONTRIBUTING.md
```diff
@@ -1,13 +1,16 @@
 # Contributing
```
pckilgore, line 1: will add comment where *not* prettier

CONTRIBUTING.md
```diff
@@ -70,53 +109,62 @@ Our pre-commit hooks verify that your commit message matches this format when co
...
-- `yarn bootstrap`: setup project by installing all dependencies and pods.
-- `yarn typescript`: type-check files with TypeScript.
-- `yarn lint`: lint files with ESLint.
-- `yarn test`: run unit tests with Jest.
-- `yarn example start`: start the Metro server for the example app.
-- `yarn example android`: run the example app on Android.
-- `yarn example ios`: run the example app on iOS.
+The `package.json` file for the root package contains various scripts for common
+tasks, and the `scripts/` folder for less-common tasks. The scripts in the
+individual packages should remain minimal.
```
pckilgore, line 126: Keeping this up to date is obviously hard (we didn't), and these are almost all self-explanitory

example/App.js
```diff
@@ -0,0 +1,6 @@
+import "react-native-gesture-handler";
+import { registerRootComponent } from "expo";
+
+import App from "./src/index";
+
+registerRootComponent(App);
```
pckilgore, line 6: There were like, four files before that essentially handled pieces of what this does now:

`__generated__/AppEntry`
`App.tsx`
`App.js`
`src/App.tsx`

Centralized here, simpler.

example/metro.config.js
```diff
@@ -0,0 +1,90 @@
+const path = require("path");
```
pckilgore, line 1: ref: https://github.com/react-navigation/react-navigation/blob/main/example/metro.config.js

example/src/index.js
```diff
@@ -1,4 +1,3 @@
-import "react-native-gesture-handler";
```
pckilgore, line 1: Now in App.js

example/webpack.config.js
```diff
@@ -0,0 +1,38 @@
+const path = require("path");
+const fs = require("fs");
+const createExpoWebpackConfigAsync = require("@expo/webpack-config");
```
pckilgore, line 3: ref: https://github.com/react-navigation/react-navigation/blob/main/example/webpack.config.js

jest/setup.js
```diff
@@ -0,0 +1 @@
+// Global setup here.
```
pckilgore, line 1: We don't really have tests yet which I didn't realize until I set this up, I assume we'll want to use this when we do, so leaving.

packages/example/App.js
```diff
@@ -1,2 +0,0 @@
-import App from "./src/App";
```
pckilgore, line 1: Not flagged as a move but this is in `/examples/App.js` now

packages/example/__generated__/AppEntry.js
```diff
@@ -1,12 +0,0 @@
-// @generated by expo-yarn-workspaces
```
pckilgore, line 1: use App.js now per above

packages/ui/package.json
```diff
@@ -88,7 +87,12 @@
     "targets": [
       "commonjs",
       "module",
-      "typescript"
+      [
+        "typescript",
+        {
+          "project": "tsconfig.build.json"
```
pckilgore, line 93: use this file to build for real, rather than resolve against monorepo versions.

packages/web/index.tsx
```diff
@@ -1,12 +0,0 @@
-/* This points directly to src/index instead of "@draftbit/ui" because we don't want to rely
```
pckilgore, line 1: now @ `web/src/index.tsx`

tsconfig.json
```diff
@@ -1,25 +1,17 @@
...
-    "scripts",
-  ],
+  "exclude": ["example/**"],
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "@draftbit/ui": [
-        "packages/react-native-jigsaw/src/index"
-      ]
+      "@draftbit/*": ["packages/*/src"]
```
pckilgore, line 6: Typescript will resolve these packages against the monorepo.  We overwrite this in each module with a `tsconfig.build.json` file so that builds for publication have real package references.

packages/types/src/index.ts
```diff
@@ -1,4 +1,5 @@
-import * as React from "react";
+import type { ComponentType, ComponentPropsWithoutRef } from "react";
+export * from "./component-types";
```
davesnx, line 2: This looks weird

davesnx (approved):

packages/types/src/index.ts
pckilgore: In what sense? It says "reexport everything from this file".  You can manually de-structure them all out if you think its clearer, I just didn't have any reason to do it.

That file is separate because the parser script expects it to look a certain way.

packages/core/src/components/HeaderMedium.tsx
```diff
@@ -17,9 +18,10 @@ type Props = {
   onPress?: () => void;
   style?: StyleProp<ViewStyle>;
   theme: typeof theme;
-};
+} & IconSlot;
 
 const HeaderMedium: React.FC<Props> = ({
+  Icon,
```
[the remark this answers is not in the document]
pckilgore: Yeah, `Icon` (the component) vs. `icon` (the string) is a little confusing, but if you look at the actual type of this module for users (after the dependency is injected), there is only one prop (icon the string).

I think there is a good reason to have the slot be the same name across all components (they implement the interface).  But if you have a better idea for the name, feel free to propose it.  It definitely isn't an `Override` anymore.  Maybe just call the prop IconSlot?

Issue #97, opened 2020-12-15
On Expo 40: Container component backgroundImage isn't backgrounded


StartupSam:
The backgroundImage property in the Container component stopped being a background image. Not immediately sure why looking at the code - any ideas on a fix?

See expected vs actual screenshots below

![IMG_4443](https://user-images.githubusercontent.com/6233577/102161987-f8a68a80-3e4d-11eb-920b-1c5e464d0b1c.PNG)
![IMG_4602](https://user-images.githubusercontent.com/6233577/102161720-728a4400-3e4d-11eb-8e43-5f941be74053.png)

dhruska:
Hey @StartupSam - try setting the `useThemeGutterPadding` prop to `false` on the Container and see if that fixes. If not, you might try using the `backgroundImageResizeMode` prop on the Container.

StartupSam:
It was already false, and was already using the backgroundImageResizeMode prop.

To fix, I just converted it to a plain ImageBackground component from react-native, but I figured you should be aware that the prop may not be working as intended.

StartupSam:
While I'm here, TypeScript question: anywhere I use Draftbit, I get the below warning message. There doesn't seem to be a types package available - is this something you have internally, or am I missing something?

```
Could not find a declaration file for module '@draftbit/ui'. '/home/samstowers/Projects/slumber-app/node_modules/@draftbit/ui/lib/commonjs/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/draftbit__ui` if it exists or add a new declaration (.d.ts) file containing `declare module '@draftbit/ui';`

```

PR #123, opened 2021-02-05
fix types location

tatchi:
This should fix https://github.com/draftbit/react-native-jigsaw/issues/97#issuecomment-745023407


peterpme (approved):
:100: thanks so much!

PR #124, opened 2021-02-05
fix aspect ratios

peterpme:


packages/react-native-jigsaw/src/components/Image.tsx
```diff
@@ -13,37 +14,86 @@ import {
...
+    };
+  }
+
+  if (aspectRatio && width && height) {
+    return {
+      aspectRatio,
+      height,
+      width: aspectRatio * height,
+    };
+  }
```
dhruska, line 39: can you eliminate this full block of code since it's covered by the next case?

peterpme: wdym?

peterpme: does it do the same thing, for sure?

dhruska: as far as i can tell you could remove lines 33 through 39 and it would work the same since when `aspectRatio && width && height` is true, `aspectRatio && height`. but if keeping it helps keep things more explicit, works for me

dhruska (approved):

packages/react-native-jigsaw/src/components/Image.tsx
dhruska: actually it's not equivalent is it

PR #127, opened 2021-02-19
updates the default placeholder text to something more useful

peterpme:
i turned of all the stupid ass props as well

### DIFF
--- a/packages/react-native-jigsaw/src/components/Text.tsx
+++ b/packages/react-native-jigsaw/src/components/Text.tsx
@@ -56,12 +56,12 @@ export const SEED_DATA = {
     children: {
       group: GROUPS.data,
       label: "Text",
-      description: "Text for the text",
+      description: "Text",
       editable: true,
       required: true,
       formType: FORM_TYPES.string,
       propType: PROP_TYPES.STRING,
-      defaultValue: "Change me under the Data Panel",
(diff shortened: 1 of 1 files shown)


packages/react-native-jigsaw/src/components/Text.tsx
```diff
@@ -56,20 +56,20 @@ export const SEED_DATA = {
     children: {
       group: GROUPS.data,
       label: "Text",
-      description: "Text for the text",
```
dhruska, line 59: lol

packages/react-native-jigsaw/src/components/Text.tsx
```diff
@@ -56,20 +56,20 @@ export const SEED_DATA = {
...
+      description: "Text",
       editable: true,
       required: true,
       formType: FORM_TYPES.string,
       propType: PROP_TYPES.STRING,
-      defaultValue: "Change me under the Data Panel",
```
dhruska, line 64: I don't recall seeing this anywhere, I wonder if this was even used?

dhruska (approved):

packages/react-native-jigsaw/src/components/Text.tsx
peterpme: never was used b/c of a change on the server lol

PR #142, opened 2021-03-23
local/remote -> image

peterpme:
replaces all localImage / remoteImage with image
replaced localImageArray with imageArray


dhruska (approved):
LGTM assuming we get image and imageArray into the server/db

PR #144, opened 2021-03-26
removes theme elevations

peterpme:
removes theme wide elevation and makes them component specific (rn paper default)

this means all components will continue to function as-is but still offers users the abillity to set different elevations on the prop level, as they used to work

image

image


packages/react-native-jigsaw/src/components/Elevation.tsx
```diff
@@ -15,16 +16,27 @@ type Props = {
   theme: typeof themeT;
 } & ViewProps;
 
-const Elevation: React.FC<Props> = ({ style, theme, ...rest }) => {
-  const flattenedStyles = StyleSheet.flatten(style) || {};
-  const { elevation } = flattenedStyles;
+const Elevation: React.FC<Props> = ({ style, theme, children, ...rest }) => {
+  const { elevation = 4, borderRadius: radius } = (StyleSheet.flatten(style) ||
+    {}) as ViewStyle;
```
pckilgore, line 21: why cast?

pckilgore (approved):
Prefer to comment or remove casts if we can, otherwise LGTM.

packages/react-native-jigsaw/src/components/Elevation.tsx
peterpme: i copied this (back) to the original implementation https://github.com/callstack/react-native-paper/blob/main/src/components/Surface.tsx#L63

we've had to do this before though so the answer is react native and typescript still don't play great together

peterpme: out of curiosity i removed the cast to see if it's been fixed and it has been, so in this case i removed it

PR #153, opened 2021-03-30
new card component

peterpme:


packages/react-native-jigsaw/src/components/Card.tsx
```diff
@@ -1,46 +1,232 @@
...
+        }}
+      >
+        <Image
+          style={{ aspectRatio }}
+          source={typeof image === "string" ? { uri: image } : image}
+          resizeMode="cover"
+        />
+        <View style={{ padding: innerPadding }}>
+          <View style={{ alignItems: textCentered ? "center" : "flex-start" }}>
+            {title || (title && title !== "") ? (
```
dhruska, line 157: Could this be simplified to this? I could be missing a case but I'm not sure when the right hand side of the `or` would hit
```suggestion
            {(title || title === "") ? (
```

peterpme: Typescript

dhruska: man typescript...

dhruska (approved):

PR #156, opened 2021-04-14
new button api

peterpme:
image


packages/react-native-jigsaw/src/components/DeprecatedButton.tsx
```diff
@@ -0,0 +1,315 @@
+import * as React from "react";
```
peterpme, line 1: the old Button.tsx

packages/react-native-jigsaw/src/components/Button.tsx
```diff
@@ -1,315 +1,204 @@
 import * as React from "react";
```
peterpme, line 1: none of this should actually be a diff, completely new file

dhruska (approved):
LGTM

PR #158, opened 2021-04-14
fix: button margins going inside elevation

pckilgore:
See: https://community.draftbit.com/c/help/button-margin-causes-background-to-be-wonky

image

### DIFF
--- a/packages/react-native-jigsaw/src/components/Button.tsx
+++ b/packages/react-native-jigsaw/src/components/Button.tsx
@@ -146,15 +146,38 @@ const Button: React.FC<Props> = ({
     },
   ];
 
+  const {
+    margin,
+    marginEnd,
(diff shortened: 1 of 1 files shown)


packages/react-native-jigsaw/src/components/Button.tsx
```diff
@@ -146,15 +146,38 @@ const Button: React.FC<Props> = ({
...
 
+  const {
+    margin,
+    marginEnd,
+    marginTop,
+    marginLeft,
+    marginRight,
+    marginBottom,
+    marginHorizontal,
+    marginVertical,
```
pckilgore, line 157: I don't even think we expose these in the builder, but I figured better safe than sorry....

peterpme (approved):

PR #170, opened 2021-04-25
Toggle Button - new component

YoussefHenna:
### Plan
**Name:** ToggleButton
**Librariess:** _None_
**Props**:
- icon: icon inside button (required) - **_string_**
- status: can be either "selected"/"unselected" - **_string_**
- onPress: to be called on press - **_function_**
- disabled: disables/enables button - **_boolean_**
- theme: for compatibility with the 'withTheme' hook. - wont be exposed to user - **_theme_**
- color: main color of button (optional, default uses theme values) - **_theme color_**
- colorSecondary: secondary color (eg: bg of unselected, text of selected), also optional default is theme color - **_theme color_**
- borderColor: color of border (defaults to theme) - **_theme color_**
- iconSize: size of the icon  - **_number_**
- size: size of the whole button (square so 1 value for width/height) - **_number_**

All props are optional, except the icon

Color props are also values of theme as in light, primary, strong etc. Not actual color hex.

I'll use theme values where applicable, like for example the disabledOpacity value of the theme.

Overall look and feel will be similar to RN paper

I can also provide the SEED_DATA, but let me know if you want me to.

No questions or concerns, seems like it will go smoothly.

### DIFF
--- a/packages/example/src/ToggleButtonExample.js
+++ b/packages/example/src/ToggleButtonExample.js
@@ -1,0 +1,34 @@
+import * as React from "react";
+import { ToggleButton } from "@draftbit/ui";
+import Section, { Container, styles } from "./Section";
+
+export default function ToggleButtonExample() {
+  const [checked, setChecked] = React.useState("unchecked");
+  const toggleChecked = () => {
+    if (checked === "unchecked") {
+      setChecked("checked");
+    } else {
+      setChecked("unchecked");
+    }
+  };
+  return (
+    <Container>
+      <Section title="ToggleButton" style={styles.row}>
+        <ToggleButton
+          style={{ margin: 5 }}
+          icon="brightness-5"
+          status={checked}
+          onPress={toggleChecked}
+        />
+        <ToggleButton
+          style={{ margin: 5 }}
+          icon="file-download"
+          width={70}
+          height={70}
+          status={checked}
+          onPress={toggleChecked}
+        />
+      </Section>
+    </Container>
+  );
+}
--- a/packages/example/src/ToggleButtonExample.js
+++ b/packages/example/src/ToggleButtonExample.js
@@ -3,13 +3,9 @@ import { ToggleButton } from "@draftbit/ui";
 import Section, { Container, styles } from "./Section";
 
 export default function ToggleButtonExample() {
-  const [checked, setChecked] = React.useState("unchecked");
-  const toggleChecked = () => {
-    if (checked === "unchecked") {
-      setChecked("checked");
-    } else {
-      setChecked("unchecked");
-    }
+  const [toggled, setToggled] = React.useState(false);
+  const toggle = () => {
+    setToggled(!toggled);
   };
   return (
     <Container>
@@ -17,16 +13,16 @@ export default function ToggleButtonExample() {
         <ToggleButton
           style={{ margin: 5 }}
           icon="brightness-5"
-          status={checked}
-          onPress={toggleChecked}
+          toggled={toggled}
+          onPress={toggle}
         />
         <ToggleButton
           style={{ margin: 5 }}
           icon="file-download"
           width={70}
           height={70}
-          status={checked}
-          onPress={toggleChecked}
+          toggled={toggled}
+          onPress={toggle}
         />
       </Section>
     </Container>
--- a/packages/react-native-jigsaw/src/components/ToggleButton.tsx
+++ b/packages/react-native-jigsaw/src/components/ToggleButton.tsx
@@ -62,6 +62,7 @@ const ToggleButton: React.FC<Props> = ({
         },
         style,
       ]}
+      {...rest}
     />
   );
 };
@@ -83,6 +84,7 @@ export const SEED_DATA = {
     icon: createIconProp({
       required: true,
     }),
+    iconSize: createIconSizeProp(),
     fieldName: createFieldNameProp({
       defaultValue: false,
       valuePropName: "toggled",
@@ -99,7 +101,6 @@ export const SEED_DATA = {
     borderColor: createColorProp({
       label: "Border Color",
     }),
-    iconSize: createIconSizeProp(),
     width: createNumberProp({
       label: "Width",
       description: "Width",
--- a/packages/react-native-jigsaw/src/components/ToggleButton.tsx
+++ b/packages/react-native-jigsaw/src/components/ToggleButton.tsx
@@ -1,0 +1,76 @@
+import * as React from "react";
+import { withTheme } from "../core/theming";
+import themeT from "../styles/DefaultTheme";
+import { colorTypes } from "../types";
+import { StyleProp, StyleSheet, ViewStyle } from "react-native";
+import Touchable from "./Touchable";
+import Icon from "./Icon";
+
+type Props = {
+  icon: string | number | { uri: string };
+  status?: "checked" | "unchecked";
+  onPress?: () => void;
+  disabled?: boolean;
+  color?: colorTypes;
+  colorSecondary?: colorTypes;
+  borderColor?: colorTypes;
+  iconSize?: number;
+  width?: number;
+  height?: number;
+  style?: StyleProp<ViewStyle>;
+  accessibilityLabel?: string;
+  theme: typeof themeT;
+};
+
+const ToggleButton: React.FC<Props> = ({
+  icon,
+  status = "unchecked",
+  onPress = () => {},
+  disabled = false,
+  color = "primary",
+  colorSecondary = "surface",
+  borderColor = "divider",
+  iconSize = 25,
+  width = 50,
+  height = 50,
+  style,
+  accessibilityLabel = "toggle button",
+  theme: { colors },
+}) => {
+  return (
+    <Touchable
+      style={[
+        styles.mainContainer,
+        {
+          width,
+          height,
+          backgroundColor:
+            status === "unchecked" ? colors[color] : colors[colorSecondary],
+          borderColor: colors[borderColor],
+        },
+        style,
+      ]}
+      onPress={onPress}
+      disabled={disabled}
+      accessibilityLabel={accessibilityLabel}
+      accessibilityState={{ disabled }}
+      accessibilityRole="button"
+    >
+      <Icon
+        name={icon}
+        size={iconSize}
+        color={status === "unchecked" ? colors[colorSecondary] : colors[color]}
+      />
+    </Touchable>
+  );
+};
+
+const styles = StyleSheet.create({
+  mainContainer: {
+    borderWidth: 1,
+    justifyContent: "center",
+    alignItems: "center",
+  },
+});
+
+export default withTheme(ToggleButton);
--- a/packages/react-native-jigsaw/src/components/ToggleButton.tsx
+++ b/packages/react-native-jigsaw/src/components/ToggleButton.tsx
@@ -3,11 +3,10 @@ import { withTheme } from "../core/theming";
 import themeT from "../styles/DefaultTheme";
 import { colorTypes } from "../types";
 import { StyleProp, StyleSheet, ViewStyle } from "react-native";
-import Touchable from "./Touchable";
-import Icon from "./Icon";
+import IconButton from "./IconButton";
 
 type Props = {
-  icon: string | number | { uri: string };
+  icon: string;
   status?: "checked" | "unchecked";
   onPress?: () => void;
   disabled?: boolean;
@@ -38,7 +37,15 @@ const ToggleButton: React.FC<Props> = ({
   theme: { colors },
 }) => {
   return (
-    <Touchable
+    <IconButton
+      icon={icon}
+      size={iconSize}
+      color={status === "unchecked" ? colors[colorSecondary] : colors[color]}
+      onPress={onPress}
+      disabled={disabled}
+      accessibilityLabel={accessibilityLabel}
+      accessibilityState={{ disabled }}
+      accessibilityRole="button"
       style={[
         styles.mainContainer,
         {
@@ -50,26 +57,13 @@ const ToggleButton: React.FC<Props> = ({
         },
         style,
       ]}
-      onPress={onPress}
-      disabled={disabled}
-      accessibilityLabel={accessibilityLabel}
-      accessibilityState={{ disabled }}
-      accessibilityRole="button"
-    >
-      <Icon
-        name={icon}
-        size={iconSize}
-        color={status === "unchecked" ? colors[colorSecondary] : colors[color]}
-      />
-    </Touchable>
+    />
   );
 };
 
 const styles = StyleSheet.create({
   mainContainer: {
     borderWidth: 1,
-    justifyContent: "center",
-    alignItems: "center",
   },
 });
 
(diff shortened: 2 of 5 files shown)


bluerssen:
@YoussefHenna Great start.
-  I think for components like this, we should default to copying everything that React Native Paper has for the component. so if there is anything there that we're missing, that's probably the easiest place to get started. Hopefully that makes the creation of the component a lot easier. 
- For the icon size, that's fine. But for the `size` (ie the rest of the button), instead of a single size parameter can we use height and width to determine the size of the button? That eliminates a config prop and lets us use the styles tab. Should automatically center the icon in the box. 
- For the colors, how will color and secondary be applied? It seems like we need 4 colors: for icon on/off and for background on/off.

YoussefHenna:
> * I think for components like this, we should default to copying everything that React Native Paper has for the component. so if there is anything there that we're missing, that's probably the easiest place to get started. Hopefully that makes the creation of the component a lot easier.

I've used [this](https://callstack.github.io/react-native-paper/toggle-button.html) as a reference, only props I've missed are accessibilityLabel and a style prop, both of which I could add. Other than that they are pretty similar

> * For the icon size, that's fine. But for the size (ie the rest of the button), instead of a single size parameter can we use height and width to determine the size of the button? That eliminates a config prop and lets us use the styles tab. Should automatically center the icon in the box.

Sure


> * For the colors, how will color and secondary be applied? It seems like we need 4 colors: for icon on/off and for background on/off.

Unselected: bg is secondary and icon is primary
Selected: bg is primary and icon is secondary

I feel this would feel natural, but I can easily add two more colors, let me know

bluerssen:
@YoussefHenna that’s fine, thanks

YoussefHenna:
Just pushed the component, tested and everything works as expected. I also added it to the example app under ToggleButton in the menu.

peterpme:
@YoussefHenna looking good so far, can you please put up a screenshot of the example and can you please make sure the type checking passes? seems like its failing below. You can test by running `yarn typescript` and `yarn build`

packages/example/src/ToggleButtonExample.js
```diff
@@ -0,0 +1,34 @@
+import * as React from "react";
+import { ToggleButton } from "@draftbit/ui";
+import Section, { Container, styles } from "./Section";
+
+export default function ToggleButtonExample() {
+  const [checked, setChecked] = React.useState("unchecked");
```
peterpme, line 6: why a string called unchecked instead of a boolean?

peterpme:
I'm struggling to understand what the difference between an IconButton and ToggleButton are? 
https://github.com/draftbit/react-native-jigsaw/blob/master/packages/react-native-jigsaw/src/components/IconButton.tsx#L48

peterpme:
I also need you to include the SEED_DATA. Check out this comment: https://github.com/draftbit/react-native-jigsaw/pull/173#issuecomment-829364338

peterpme:
@YoussefHenna Great start! General feedback for components moving forward:

- Can you please include a screenshot/gif? If there's a active state or w/e, include it
- SEED_DATA needs to be included. Check out this comment: https://github.com/draftbit/react-native-jigsaw/pull/173#issuecomment-829364338 to explain how. It can't be an object anymore, it must be a function. That explains how
- Icons are a little finnicky within our system. You need to pass down an Icon as you can see here: https://github.com/draftbit/react-native-jigsaw/blob/master/packages/react-native-jigsaw/src/components/IconButton.tsx#L51
- Icon names also must be prefixed with the library name, otherwise it will be Material. So `FontAwesome/photo` instead of just `photo`. Or `MaterialIcons/photo`



Feedback for this particular component:

- I think this is just an IconButton that might be wrapped in something. React Native Paper also wraps this in an IconButton and includes a Context / Provider pattern. Any reason why you didn't include that context / provider pattern the way they did?

packages/example/src/ToggleButtonExample.js
YoussefHenna: That's how RN Paper does it. Check it [here](https://callstack.github.io/react-native-paper/toggle-button.html). As requested I followed how they did it. Boolean definitely makes more sense. Let me if I should change it.

YoussefHenna:
> * Can you please include a screenshot/gif? If there's a active state or w/e, include it

ezgif-1-8ea8fd6abd54

> * SEED_DATA needs to be included. Check out this comment: #173 (comment) to explain how. It can't be an object anymore, it must be a function. That explains how

I'll add it





> * Icons are a little finnicky within our system. You need to pass down an Icon as you can see here: https://github.com/draftbit/react-native-jigsaw/blob/master/packages/react-native-jigsaw/src/components/IconButton.tsx#L51
Icon names also must be prefixed with the library name, otherwise it will be Material. So FontAwesome/photo instead of just photo. Or MaterialIcons/photo



>* I think this is just an IconButton that might be wrapped in something. React Native Paper also wraps this in an IconButton

I'll use IconButton instead of icon, and pass the icon prop to it. That way it can handle the icons and everything should work as normal. 



>* includes a Context / Provider pattern. Any reason why you didn't include that context / provider pattern the way they did?

This would only be useful when implementing the Group component. Because all the other group components didn't use that pattern. Thought I'd match the ones in jigsaw rather than RN Paper. 




>* I think you should be using a boolean for the toggle, unless you have a very good reason why this isn't a toggle boolean.

That's how RN Paper does it. Check it [here](https://callstack.github.io/react-native-paper/toggle-button.html). As requested I followed how they did it. Boolean definitely makes more sense. Let me know if I should change it.

YoussefHenna:
I pushed the SEED_DATA but need you to review it as I am unsure if it's all done correctly or not.

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  accessibilityLabel?: string;
+  theme: typeof themeT;
+};
+
+const ToggleButton: React.FC<Props> = ({
+  icon,
+  status = "unchecked",
+  onPress = () => {},
+  disabled = false,
+  color = "primary",
```
peterpme, line 42: iconColor

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  colorSecondary = "surface",
```
peterpme, line 43: backgroundColor

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  borderColor = "divider",
+  iconSize = 25,
+  width = 50,
+  height = 50,
+  style,
+  accessibilityLabel = "toggle button",
```
peterpme, line 49:
```suggestion
  accessibilityLabel = "ToggleButton",
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  status = "unchecked",
+  onPress = () => {},
+  disabled = false,
+  color = "primary",
+  colorSecondary = "surface",
+  borderColor = "divider",
+  iconSize = 25,
+  width = 50,
+  height = 50,
+  style,
```
peterpme, line 48: style should go last, after theme

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  accessibilityLabel = "toggle button",
+  theme: { colors },
```
peterpme, line 50: add a `...rest` props here

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+      label: "Width",
+      description: "Width",
+      defaultValue: 50,
+    }),
+    height: createNumberProp({
+      label: "Height",
+      description: "Height",
+      defaultValue: 50,
+    }),
+    accessibilityLabel: createTextProp({
```
peterpme, line 133: I imagine we'll re-use this a lot. Can you add a function inside component-types, just like the rest of these:

```
accessibilityLabel: createAccessibilityLabel()
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+      description: "Whether the button should be disabled",
+    }),
+    color: createColorProp(),
+    colorSecondary: createColorProp({
+      label: "Secondary Color",
+    }),
+    borderColor: createColorProp({
+      label: "Border Color",
+    }),
+    iconSize: {
```
peterpme, line 112: lets create a function called `createIconSizeProp()` within component-types.ts since we use this in a couple places

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+      valuePropName: "status",
+      handlerPropName: "onPress",
+    },
+    onPress: createActionProp(),
+    disabled: createBoolProp({
+      label: "Disabled",
+      description: "Whether the button should be disabled",
+    }),
+    color: createColorProp(),
+    colorSecondary: createColorProp({
```
peterpme, line 106:
```suggestion
    backgroundColor: createColorProp({
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+      defaultValue: "unchecked",
+      valuePropName: "status",
+      handlerPropName: "onPress",
+    },
+    onPress: createActionProp(),
+    disabled: createBoolProp({
+      label: "Disabled",
+      description: "Whether the button should be disabled",
+    }),
+    color: createColorProp(),
```
peterpme, line 105:
```suggestion
    iconColor: createColorProp(),
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
+import * as React from "react";
+import { withTheme } from "../core/theming";
+import themeT from "../styles/DefaultTheme";
```
peterpme, line 3:
```suggestion
import type { Theme } from '../styles/DefaultTheme'
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  disabled?: boolean;
+  color?: colorTypes;
+  colorSecondary?: colorTypes;
+  borderColor?: colorTypes;
+  iconSize?: number;
+  width?: number;
+  height?: number;
+  style?: StyleProp<ViewStyle>;
+  accessibilityLabel?: string;
+  theme: typeof themeT;
```
peterpme, line 34:
```suggestion
  theme: Theme
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  FIELD_NAME,
+} from "../core/component-types";
+
+type Props = {
+  icon: string;
+  status?: "checked" | "unchecked";
+  onPress?: () => void;
+  disabled?: boolean;
+  color?: colorTypes;
+  colorSecondary?: colorTypes;
```
peterpme, line 27:
```suggestion
  backgroundColor?: colorTypes;
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  createActionProp,
+  FIELD_NAME,
+} from "../core/component-types";
+
+type Props = {
+  icon: string;
+  status?: "checked" | "unchecked";
+  onPress?: () => void;
+  disabled?: boolean;
+  color?: colorTypes;
```
peterpme, line 26:
```suggestion
  iconColor?: colorTypes;
```

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+  theme: { colors },
+}) => {
+  return (
+    <IconButton
+      icon={icon}
+      size={iconSize}
+      color={status === "unchecked" ? colors[colorSecondary] : colors[color]}
+      onPress={onPress}
+      disabled={disabled}
+      accessibilityLabel={accessibilityLabel}
```
peterpme, line 59: you can remove these 3 accessibility props because an IconButton uses the new Pressable api which doesn't support these

packages/react-native-jigsaw/src/components/ToggleButton.tsx
peterpme, line 133: you can remove this since its not supported by the pressable api

YoussefHenna:
Ignore this, just a reminder for myself:

Change status to use boolean instead of string

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+export const SEED_DATA = {
+  name: "Toggle Button",
+  tag: "ToggleButton",
+  category: COMPONENT_TYPES.button,
+  layout: {},
+  props: {
+    icon: createIconProp({
+      required: true,
+    }),
+    fieldName: {
```
peterpme, line 94:
```suggestion
    fieldName: createFieldName({
      defaultValue:  false,
      valuePropName: "toggled",
    })
```
same as the rest of them, would be nice if we didn't have to import any of the objects and just have functions

you can make the default handlerPropName onPress so you don't have to include it

packages/react-native-jigsaw/src/components/ToggleButton.tsx
```diff
@@ -0,0 +1,114 @@
...
+      style={[
+        styles.mainContainer,
+        {
+          width,
+          height,
+          backgroundColor: toggled ? colors[colorSecondary] : colors[color],
+          borderColor: colors[borderColor],
+        },
+        style,
+      ]}
```
peterpme, line 64: `rest` is defined but not used: still need to add it down here (into IconButton):

```<IconButton allotherprops {...rest} />```

File: packages/react-native-jigsaw/src/core/component-types.ts [hunk 18]
```diff
@@ -252,6 +252,25 @@ export const createColorProp = (overrides = {}) => ({
...
+  editable: true,
+  required: false,
+  formType: FORM_TYPES.flatArray,
+  propType: PROP_TYPES.NUMBER,
+  defaultValue: 32,
+  options: [12, 16, 24, 32, 48, 64],
+  ...overrides,
+});
+
+export const createFieldNameProp = (overrides = {}) => ({
```
peterpme, line 268: great job!

YoussefHenna:
@peterpme 

Done.

peterpme:
Great job!!

PR #172, opened 2021-04-26
fix(button): expose text align on button varieties

dhruska:
Allows user to set text align property on button/link components, and sets seed data defaults

Example: https://snack.expo.io/@donald/(draftbit)-basic-profile


peterpme (approved):
wow great fix!

PR #173, opened 2021-04-27
Radio Button - New Component

IsmaFunes:
## Description

This component will const of 3 parts, based on `react-native-paper` implementation:

- RadioButton: Component used when you need an isolated radio button
- RadioButton.Group: Component for a group of radio buttons, only allowing for one to be currently selected
- RadioButton.Row: Used inside RadioButton.Group

### Radio Button

**selected**
Type: boolean
Whether radio is selected.

**disabled**
Type: boolean
Whether radio is disabled.

**onPress**
Type: () => void
Function to execute on press.

**uncheckedColor**
Type: string
Custom color for unchecked radio.

**color**
Type: string
Custom color for radio.

**theme**
Type: inherited from withTheme HOC

**style**
Type: ViewStyle
Styles the container of the radio button

**selectedIcon**
Type: string
Icon to show when the radio button is selected
default: "MaterialIcons/radio-button-checked"

**unselectedIcon**
Type: string
Icon to show when the radio button is selected
default: "MaterialIcons/radio-button-unchecked"

**size**
Type: number
Size of the icon
default: `Config.radioButtonSize` (24)

-----

### RadioButton.Row

**value (required)**
Type: string
Value of the radio button

**label (required)**
Type: string | ReactNode
Label to show along with the radio button. It can be fully customized if passed a ReactNode

**direction**
Type: 'row' | 'row-reverse'
Specifies the flex direction of the container, allowing the icon to appear on the left or on the right.
default: 'row'

**labelContainerStyle**
Type: StyleProp<ViewStyle>
Styles to pass to the customize the container of the label

**radioButtonStyle**
Type: StyleProp<ViewStyle>
Styles to pass to the customize the radio button itself. It will be passed down to the RadioButton.

**onPress**
Type: (value: string) => void
Unlike the onPress property of the radio button, this callback passes its value to the caller in order to know which radio button has been just selected.

This component also receives the same props as `RadioButton`.

----

### RadioButton.Group

**onValueChange (required)**
Type: (value: string) => void
Function to be executed on selection change.

**value (required)**
Type: string
Value of the currently selected radio button.

**children (required)**
Type: ReactNode
Contains one or many RadioButton.Row

**containerStyle**
Type: StyleProp<ViewStyle>
Styles the container of the whole component.

**direction**
Type: "horizontal" | "vertical"
Determines if the radio buttons inside of the group will align horizontally or vertically.

## Notes

- The `RadioButton.Group` only works if passed a `value`. It has no inner state. This could be changed in order to allow the component to be uncontrolled and stateful if no value has been provided beforehand, otherwise it should use the provided one

## Example

Here's an expo snack with an example of the component implementation. It's not styled nor finished, just a taste of how would I implement it.

https://snack.expo.io/@maelfunes/radiobuttongroup


bluerssen:
Hey Ismael -- Thanks for this, great start.

Some thoughts:

Radio Button
- We have an existing Radio Button component, we should get that one working / fixed / displaying properly on web and use it here. You can update it to include things above, but also will need to consider our theme colors, etc. 
- You should review our theme attributes how we use them in other components, we'll want to leverage those for colors. 
- Unlike RN Paper, I think we want the traditional radio button circle / filled circle instead of the check box. 

Radio Button Row
- This is more than just the radio button, this is a full-width row that includes the radio button on the right or left side and text on the other side. So we need to be able to change the typography, style it, adding margins/padding, etc. Essentially a view with props, with a text component and the radio button, and ability to set all those props. See https://airtable.com/tblOyHD3gpuQIqhrn/viwtCHLCj3Z0CIwQv/recZHSvg83KsTmNPz?blocks=hide

Radio Button Group
- Make sure we can set the styles for this Group like we could any other view. 

@peterpme What is the right way for Ismael to set up the onPress actions for Radio Button Rows / Groups etc. Selecting one should deselect the others, but how do we make those values available in the builder from a data perspective?

IsmaFunes:
> Hey Ismael -- Thanks for this, great start.
> 
> Some thoughts:
> 
> Radio Button
> 
> * We have an existing Radio Button component, we should get that one working / fixed / displaying properly on web and use it here. You can update it to include things above, but also will need to consider our theme colors, etc.
> * You should review our theme attributes how we use them in other components, we'll want to leverage those for colors.
> * Unlike RN Paper, I think we want the traditional radio button circle / filled circle instead of the check box.
> 
> Radio Button Row
> 
> * This is more than just the radio button, this is a full-width row that includes the radio button on the right or left side and text on the other side. So we need to be able to change the typography, style it, adding margins/padding, etc. Essentially a view with props, with a text component and the radio button, and ability to set all those props. See https://airtable.com/tblOyHD3gpuQIqhrn/viwtCHLCj3Z0CIwQv/recZHSvg83KsTmNPz?blocks=hide
> 
> Radio Button Group
> 
> * Make sure we can set the styles for this Group like we could any other view.
> 
> @peterpme What is the right way for Ismael to set up the onPress actions for Radio Button Rows / Groups etc. Selecting one should deselect the others, but how do we make those values available in the builder from a data perspective?

Ok, I'll fix the existing one and reuse it inside the Row. Some more thoughts/questions on my end:

- Should the "lone" Radio Button be label-less and only include the button? (I think this is what rn-paper does). If that is the case, then some of the props would go into the Row and not directly into the Radio Button itself.
- As for the RadioButton.Row, we could add styles for `container`, `label` and `icon` (which I think will be the three main parts of the component), hence providing the capability to fully customize the style of the row.
- RadioButtonGroup could have a `containerStyle` or something alike, in order to style the View that englobes all the Rows.

peterpme:
Here's the current radio button group component we have, it should just need minor adjustments.

https://github.com/draftbit/react-native-jigsaw/blob/master/packages/react-native-jigsaw/src/components/RadioButtonGroup.tsx#L143

The line points to how the seed data works inside draftbit. In this case, all the correct props should be there. If there are some that need to be renamed / added, you should be able to figure it out.

In order to make the props more consistent, we're moving away from using objects and using functions instead. For example, take a look at https://github.com/draftbit/react-native-jigsaw/blob/master/packages/react-native-jigsaw/src/components/Button.tsx#L231-L239

Sometimes you need to override the defaults. In this case we override label, description for disabled/loading. the onPress handler doesn't need any overrides so it just takes an empty function.

Please go with this style moving forward. In the RadioButtonGroup case, please convert the old style object to using this new style:

```
  disabled: createBoolProp({
    label: "Disabled",
    description: "Whether the button should be disabled",
  }),
  loading: createBoolProp({
    label: "Loading",
    description: "Whether to show a loading indicator",
  }),
  onPress: createActionProp(),
  ```

**Comment by IsmaFunes** [comment]
RadioButtonGroup

Attached examples GIF (Sorry FPS are a bit low)

File: packages/react-native-jigsaw/src/components/RadioButton/index.tsx [hunk 1]
```diff
@@ -0,0 +1,18 @@
+import RadioButtonGroup, { RadioButtonGroupProps } from "./RadioButtonGroup";
+import RadioButtonRow, { RadioButtonRowProps } from "./RadioButtonRow";
+import { RadioButtonProps, default as SingleRadioButton } from "./RadioButton";
+import React from "react";
```
peterpme, line 4: place this at the top

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -0,0 +1,156 @@
...
+} from "../../core/component-types";
+
+export interface RadioButtonRowProps extends Omit<RadioButtonProps, "onPress"> {
+  label: string | React.ReactNode;
+  value: string;
+  labelContainerStyle: StyleProp<ViewStyle>;
+  radioButtonStyle?: StyleProp<ViewStyle>;
+  labelStyle?: StyleProp<TextStyle>;
+  onPress?: (value: string) => void;
+  direction?: "row" | "row-reverse";
```
peterpme, line 29: instead of making this a string, can you make this an enum? Row | RowReverse?

you can make it so the enum translates to a lowercase row or row-reverse

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -0,0 +1,156 @@
...
+  ViewStyle,
+  StyleSheet,
+  TouchableOpacity,
+  TextStyle,
+} from "react-native";
+import RadioButton, {
+  SEED_DATA as RADIO_BUTTON_SEED_DATA,
+  RadioButtonProps,
+} from "./RadioButton";
+import { View } from "react-native";
```
peterpme, line 13: can you consolidate the imports? You're already importing `react-native`

packages/react-native-jigsaw/src/components/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+  COMPONENT_TYPES,
+  createBoolProp,
+  createColorProp,
+  createActionProp,
+  createNumberProp,
+  createIconProp,
+} from "../../core/component-types";
+import Config from "../Config";
+import IconButton from "../IconButton";
+import themeT from "../../styles/DefaultTheme";
```
peterpme, line 15:
```suggestion
import type { Theme } from "../../styles/DefaultTheme";
```

packages/react-native-jigsaw/src/components/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+const RadioButton: React.FC<RadioButtonProps> = ({
+  selected,
+  disabled = false,
+  color,
+  unselectedColor,
+  onPress = () => {},
+  style,
+  size = Config.radioButtonSize,
+  selectedIcon = "MaterialIcons/radio-button-checked",
+  unselectedIcon = "MaterialIcons/radio-button-unchecked",
```
peterpme, line 39: Can you add `...rest` here as a prop, then spread it as the last prop?

packages/react-native-jigsaw/src/components/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+}) => {
+  return (
+    <IconButton
+      style={style}
+      icon={selected ? selectedIcon : unselectedIcon}
+      color={selected ? color : unselectedColor}
+      disabled={disabled}
+      onPress={onPress}
+      size={size}
```
peterpme, line 48: add {...rest} here as the last one. then put style right above it

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonFieldGroup.tsx
```diff
@@ -0,0 +1,42 @@
...
+  style?: StyleProp<ViewStyle>;
+}
+
+const RadioButtonFieldGroup: React.FC<Props> = ({
+  label,
+  children,
+  theme,
+  labelStyle,
+  style,
+  ...other
```
peterpme, line 20:
```suggestion
  ...rest
```

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,78 @@
...
+  FORM_TYPES,
+  FIELD_NAME,
+  createTextProp,
+  createTextStyle,
+} from "../../core/component-types";
+import themeT from "../../styles/DefaultTheme";
+import { radioButtonGroupContext } from "./context";
+
+export interface RadioButtonGroupProps {
+  direction?: "horizontal" | "vertical";
```
peterpme, line 14: can you make this an enum? You can do this:

```
enum Direction {
Horizontal = "horizontal",
Vertical = "vertical"
}
```

and i believe it will convert that string correctly to a lower case string

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,78 @@
...
+  style?: StyleProp<ViewStyle>;
+  value: string;
+  onValueChange?: (value: string) => void;
+  theme: typeof themeT;
```
peterpme, line 18:
```suggestion
  theme: Theme
```

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from "react";
+import { View, StyleProp, ViewStyle } from "react-native";
+import {
+  COMPONENT_TYPES,
+  FORM_TYPES,
+  FIELD_NAME,
+  createTextProp,
+  createTextStyle,
+} from "../../core/component-types";
+import themeT from "../../styles/DefaultTheme";
```
peterpme, line 10:
```suggestion
import type { Theme } from "../../styles/DefaultTheme";
```

packages/example/src/RadioButtonExample.js
```diff
@@ -1,86 +1,151 @@
...
+import { withTheme } from "@draftbit/ui";
 
-export default function RadioButtonGroupExample() {
-  const [selected, onSelect] = React.useState("First");
+const LoneRadioButtonWrapper = ({
+  label,
+  children
+}) => (
+  <View style={{
+    flex: 1,
```
peterpme, line 12: can you place these styles in StyleSheet.create?

peterpme:
Great start! A bunch of comments. Let me know :) 

Can you make the `unselected` checkbox not include the check? Seems like it should be an empty box instead of one with a checkmark in it, don't you agree?

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -0,0 +1,156 @@
...
+  ...other
+}) => {
+  const {
+    value: contextValue,
+    onValueChange,
+    direction: parentDirection,
+  } = useRadioButtonGroupContext();
+
+  const handlePress = () => {
+    onPress(value);
```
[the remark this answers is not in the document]
IsmaFunes: No. onPress is initialized as an empty function

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { radioButtonGroupContext } from "./context";
+
+export interface RadioButtonGroupProps {
+  direction?: "horizontal" | "vertical";
+  style?: StyleProp<ViewStyle>;
+  value: string;
+  onValueChange?: (value: string) => void;
```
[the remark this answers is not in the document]
IsmaFunes: Shouldn't be. My bad

packages/example/src/RadioButtonExample.js
```diff
@@ -1,86 +1,155 @@
 import * as React from "react";
-import { RadioButtonGroup, RadioButtonFieldGroup } from "@draftbit/ui";
+import { Icon, RadioButton, Row, RadioButtonFieldGroup, withTheme } from "@draftbit/ui";
 import Section, { Container } from "./Section";
+import { Text, View, StyleSheet } from "react-native";
 
-export default function RadioButtonGroupExample() {
-  const [selected, onSelect] = React.useState("First");
+const LoneRadioButtonWrapper = ({
```
peterpme, line 6:
```suggestion
const SingleRadioButtonWrapper = ({
```

packages/example/src/RadioButtonExample.js
```diff
@@ -1,86 +1,155 @@
...
+
+const RadioButtonGroupExample = ({
+  theme
+}) => {
+  const [selected, onSelect] = React.useState("1");
   const handleSelect = (value) => onSelect(value);
   return (
     <Container>
+
+      <Section title="Lone radio buttons">
```
peterpme, line 29:
```suggestion
      <Section title="Single Radio Buttons">
```

packages/react-native-jigsaw/src/components/RadioButton/RadioButtonFieldGroup.tsx
```diff
@@ -0,0 +1,42 @@
...
+  labelStyle,
+  style,
+  ...rest
+}) => {
+  return (
+    <View style={style}>
+      <Text
+        style={[
+          {
+            fontSize: theme.typography.headline4.fontSize,
```
peterpme, line 27: im not sure what to do with these theme values b/c they shouldn't exist lol

peterpme:
Super close! You have a bunch of linting errors left. You can run this locally via `yarn lint` or `yarn typescript` (its in package.json) or you can take a look at the failing checks in the PR

peterpme:
Im gonna merge this in for now so we can test it. Great job!!!

PR #192, opened 2021-05-18
Checkbox Group - New Component

IsmaFunes:
# Description

In my opinion, this component should be very similar to Radio Button Group in order to keep consistency between them.

This component will const of 3 parts:

- Checkbox: Isolated checkbox (no label, just a check/uncheck box)
- Checkbox.Group: Component for a group of checkboxes, allowing none or many to be selected.
- Checkbox.Row: Used inside Checkbox.Group. It is a checkbox along with a label. Tapping the whole row triggers the change.

Checkbox.Row and Checkbox.Group will communicate via context API.

# Props

### Exposed to the user

**selected**
Type: boolean
Whether the checkbox is selected.

**disabled**
Type: boolean
Whether the checkbox is disabled.

**onPress**
Type: () => void
Function to execute on press.

**color**
Type: string
Custom color for the checkbox.

**uncheckedColor**
Type: string
Custom color for unchecked checkbox. If not provided, `color` will be used.

**style**
Type: ViewStyle
Styles the checkbox as a view

**checkedIcon**
Type: string
Icon to show when the checkbox is selected

**uncheckedIcon**
Type: string
Icon to show when the checkbox is selected

**size**
Type: number
Size of the icon
default: 24 (could be added to Config object)

-----

### Exposed to the user

**value (required)**
Type: string
Value of the checkbox

**label (required)**
Type: string
Label to show along with the checkbox. 

**direction**
Type: 'row' | 'row-reverse'
Specifies the flex direction of the container, allowing the icon to appear on the left or on the right.
default: 'row'

**onPress**
Type: (value: string) => void
Unlike the onPress property of the Checkbox, this callback passes its value to the caller in order to know which checkbox has been just selected.

**selected**
Type: boolean
If provider, this particular row will always be selected, ignoring the values of the group it belongs to.

### Not exposed (dev only)

**labelContainerStyle**
Type: StyleProp<ViewStyle>
Styles to pass to the customize the container of the label

**checkboxStyle**
Type: StyleProp<ViewStyle>
Styles to pass to the customize the checkbox itself. It will be passed down to the Checkbox component.

**label**
Type: ReactNode
Asides from being able to pass a string to the label, devs will also be able to pass a react node to it.

This component also receives the same props as `Checkbox`.

----

### Exposed to the user

**values (required)**
Type: string[]
Array of currently selected checkbox values.

**onValueChange (required)**
Type: (values: string[]) => void
Function to be executed on selection change.

**style**
Type: StyleProp<ViewStyle>
Styles the container of the whole component.

**direction**
Type: "horizontal" | "vertical"
Determines if the checkboxes inside of the group will align horizontally or vertically.

## Notes

- _(question)_ Currently, the checkbox possible status are `"checked" | "indeterminate" | "unchecked"`. Should we keep this behavior? Does "indeterminate" means that the checkbox is not selected neither unselected?
- _(question)_ Similar to react-native-paper, there is a CheckboxAndroid and a CheckboxIOS components. Should these be replaced?
- In order to be consistent between Radio Button Group and Checkbox Group, I added the same `direction` prop.
- Same as above with `selected`, which could be changed to `checked`


peterpme:
Great start! You're right, RadioButton + Checkbox are very similar. I think you can even re-use your RadioButton work for Checkbox.

The only difference is that a checkbox can select multiple things while a RadioButton can only select one.

IsmaFunes:
> Great start! You're right, RadioButton + Checkbox are very similar. I think you can even re-use your RadioButton work for Checkbox.
> 
> The only difference is that a checkbox can select multiple things while a RadioButton can only select one. With that in mind, you can get started!

Great! Thanks for the review.

How about the `indeterminate` status and the IOS and Android components though?

bluerssen:
My understanding of checkboxes is that indeterminate is used for aggregating nested values, etc. Usually the visual is a minus sign (-). Although we wont support aggregating nested values, we could use the indeterminate status for invalid values passed to the checkbox. We should support with a "-". 

We do not want iOS and Android versions, so lets unify around the version with the box, the android version in react native paper. Thank you!

IsmaFunes:
https://user-images.githubusercontent.com/33071292/118908052-07aad080-b8f7-11eb-84c4-12bf88a377bd.mp4

packages/example/src/CheckboxExample.js
```diff
@@ -0,0 +1,120 @@
...
+    <View>{children}</View>
+  </View>
+);
+
+const CheckboxExample = ({ theme }) => {
+  const [checked, setChecked] = React.useState(true);
+  const [selectedValues, setSelectedValues] = React.useState([]);
+
+  const handleValueSelected = (value, selected) => {
+    console.log(value, selected);
```
peterpme, line 26: remove console

peterpme (approved):

PR #210, opened 2021-05-25
New Component- WebView

IsmaFunes:
WebView component implementation

### Notes:

- On web, many URIs won't work because of the `X-Frame-Options: SAMEORIGIN` rule
- There's a known issue with scrolling on Android that is solved by injecting javascript into the WebView. This makes the webview height adjust to its content.


example/src/WebViewExample.js
```diff
@@ -0,0 +1,36 @@
...
+  <Container>
+    <Section title="With URI">
+      <WebView
+        style={style.webView}
+        source={{
+          uri: "https://docs.expo.io/",
+        }}
+      />
+    </Section>
+    <Section title="With custom HTML">
```
peterpme, line 16: custom html won't work for our platform, you can remove it

packages/native/src/components/WebView.tsx
```diff
@@ -0,0 +1,66 @@
...
+        onMessage={onMessage}
+      />
+    </ScrollView>
+  );
+};
+
+const BrowserWebView: React.FC<Props> = ({ source, style }) =>
+  React.createElement("iframe", {
+    style: StyleSheet.flatten(style),
+    height: style?.height,
```
peterpme, line 45: does this technique actually work? i think you need to call StyleSheet.flatten no?

packages/native/src/components/WebView.tsx
```diff
@@ -0,0 +1,66 @@
...
+        injectedJavaScript={injectFirst}
+        onMessage={onMessage}
+      />
+    </ScrollView>
+  );
+};
+
+const BrowserWebView: React.FC<Props> = ({ source, style }) =>
+  React.createElement("iframe", {
+    style: StyleSheet.flatten(style),
```
peterpme, line 44: do you need to call StyleSheet.flatten here?

packages/native/src/components/WebView.tsx
IsmaFunes: You're right. My bad

packages/native/src/components/WebView.tsx
IsmaFunes: Yes. The element receives and object for styles, but RN styles are actually numbers representing an object. Flatten returns the actual object, otherwise it will throw an error

example/src/WebViewExample.js
IsmaFunes: Do you mean the implementation or the example? (or both)

peterpme: I suppose for the sake of example we can leave it!

peterpme:
Great job!

PR #227, opened 2021-06-10
various jigsaw fixes

peterpme:
- allows you to update individual files, ie: `yarn upload:run StarRating.tsx View.js`
- creates static prop generation to separate different fields

### DIFF
--- a/scripts/collector.js
+++ b/scripts/collector.js
@@ -21,33 +21,47 @@ const IGNORED_FILES = [
 const ERROR_FILES = [];
 const COMPLETED_FILES = [];
 
-async function main() {
+async function main(list) {
   console.log("Running on", getUrl());
+
+  if (list) {
+    console.log("ONLY running", JSON.stringify(list));
+  }
 
   const nativeFiles = await globAsync(`${NATIVE_PATH}/**/*.tsx`);
   const componentFiles = await globAsync(`${COMPONENT_PATH}/**/*.tsx`);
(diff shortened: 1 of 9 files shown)


scripts/collector.js
```diff
@@ -21,34 +21,48 @@ const IGNORED_FILES = [
 const ERROR_FILES = [];
 const COMPLETED_FILES = [];
 
-async function main() {
+async function main(list) {
   console.log("Running on", getUrl());
 
+  if (list) {
+    console.log("ONLY running", JSON.stringify(list));
```
adnelson, line 28: 💣

adnelson: ah nm, logging is fine here

scripts/collector.js
```diff
@@ -21,34 +21,48 @@ const IGNORED_FILES = [
...
+        ...componentFiles,
+        ...screenFiles,
+        ...mappingFiles,
+      ].filter((file) => {
+        const name = file.split("/").pop();
+
+        if (
+          name.includes("web") ||
+          name.includes("ios") ||
+          name.includes("android")
```
adnelson, line 57: can we be more specific than includes?

peterpme: yes/no potentialy - some of the names are DatePicker.web.tsx or DatePicker.web.js, etc. in React Native web, ios, android extensions will load that file on that specific platform so i figured this would be ok, lmk

PR #229, opened 2021-06-10
tushar/datepicker updates

tush13ar:
Date picker updated with requested changes.


example/src/DatePickerExample.js
```diff
@@ -10,22 +10,25 @@ function DatePickerExample({ theme }) {
     <Container style={{ backgroundColor: theme.colors.background }}>
       <Section title="Underline">
         <DatePicker
-          label="Date"
+          label="Date (without date prop)"
```
peterpme, line 13: i think you'll need at least a date always, right?

adnelson (approved):

PR #239, opened 2021-06-23
fix(radiobutton): fix radiobutton seed data and make plain button work

dhruska:


packages/core/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -58,16 +57,10 @@ export const SEED_DATA = {
...
     direction: createDirectionProp(),
-    value: createTextProp({
-      label: "value",
-      description: "Currently selected value of the radio button group",
-      defaultValue: null,
-      required: true,
-    }),
     fieldName: createFieldNameProp({
       handlerPropName: "onValueChange",
-      valuePropName: "radioButtonGroupValue",
```
dhruska, line 69: This was creating a prop called `radioButtonGroupValue` which is not consumed by this component

packages/core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -3,24 +3,25 @@ import { StyleProp, ViewStyle } from "react-native";
```
dhruska, line 3: It's not clear to me what purpose the plain RadioButton component serves - I think we should make it work inside the group (what this PR does), or just remove it so folks don't get confused as easily.

peterpme (approved):

PR #240, opened 2021-06-24
Additional radiobutton fixes

dhruska:


packages/core/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -60,7 +60,7 @@ export const SEED_DATA = {
     fieldName: createFieldNameProp({
       handlerPropName: "onValueChange",
       valuePropName: "value",
-      defaultValue: "value",
+      defaultValue: "radioButtonGroupValue",
```
dhruska, line 63: Renders the following:
```js
const [radioButtonGroupValue, setRadioButtonGroupValue] = React.useState(
  undefined
);
```

packages/core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -52,13 +54,14 @@ const RadioButton: React.FC<RadioButtonProps> = ({
     }
   };
 
-  const selected = contextValue === value;
+  const isSelected =
+    selected || (contextValue != null && contextValue === value);
```
dhruska, line 58: RadioButton is utilized by RadioButtonRow, which passes down a `selected` prop. RadioButtons can also be used standalone inside a RadioButtonGroup, hence the need for also reading `contextValue`.

peterpme (approved):

PR #247, opened 2021-06-29
Add ability to change textfield/picker bg and border

dhruska:


packages/core/src/components/TextField.tsx
```diff
@@ -398,23 +398,56 @@ class TextField extends React.Component<Props> {
...
         {leftIconName && leftIconMode === "outset" ? (
           <Icon {...leftIconProps} style={leftIconStyle} />
         ) : null}
-        <View style={[containerStyle, style ? { height: style.height } : {}]}>
+        <View
+          style={[
+            containerStyle,
+            style
+              ? {
+                  ...(style.height ? { height: style.height } : {}),
```
peterpme, line 424: have you tried

```
height: style.height,
height: style.height ?? undefined,
height: style.height || undefined
height: style.height ? style.height : undefined // this one for sure works
```

dhruska: Yeah - surprisingly, those will not work. See this Snack - `undefined` overrides an explicitly set value.

https://snack.expo.io/@donald/groaning-edamame

peterpme (approved):

PR #250, opened 2021-06-29
Add disabled to all textinputs

dhruska:


packages/core/src/mappings/TextInput.js
```diff
@@ -73,17 +73,6 @@ export const SEED_DATA = {
       formType: FORM_TYPES.boolean,
       propType: PROP_TYPES.BOOLEAN,
     },
-    keyboardAppearance: {
```
dhruska, line 76: This is already in TEXT_INPUT_PROPS

peterpme (approved):

PR #255, opened 2021-06-30
Add initialValue functionality to all form components

dhruska:
This is a prop that can be used to set the initial value of the component. This is useful because sometimes components are loaded in Draftbit within a render prop or FlatList, where a useEffect can not be utilized within the parent component.

### DIFF
--- a/example/src/CheckboxExample.js
+++ b/example/src/CheckboxExample.js
@@ -21,6 +21,7 @@ const SingleCheckboxWrapper = ({ label, children }) => (
 
 const CheckboxExample = ({ theme }) => {
   const [checked, setChecked] = React.useState(true);
+  const [checked2, setChecked2] = React.useState(true);
   const [selectedValues, setSelectedValues] = React.useState([]);
 
   const handleValueSelected = (value, selected) => {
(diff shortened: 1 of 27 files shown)


peterpme (approved):

example/src/CheckboxExample.js
```diff
@@ -33,7 +34,7 @@ const CheckboxExample = ({ theme }) => {
     }
   };
 
-  const handlePress = () => setChecked((prevState) => !prevState);
+  const handlePress = (value) => setChecked(() => value);
```
peterpme, line 37: shouldn't this be `setChecked(value)`?

example/src/DatePickerExample.js
```diff
@@ -85,6 +89,15 @@ function DatePickerExample({ theme }) {
           date={date}
           onDateChange={handleChange}
         />
+
+        <DatePicker
+          label="Date with initial value"
+          placeholder="Select a date..."
+          type="solid"
+          date={date2}
+          onDateChange={(date) => setDate2(date)}
```
peterpme, line 98:
```suggestion
          onDateChange={(d) => setDate2(d)}
```

example/src/CheckboxExample.js
dhruska: Yes! good catch

peterpme (approved):

PR #268, opened 2021-07-22
Improvements to TextField and Picker components

dhruska:
- Allow text style props
- Add `placeholderTextColor` prop to Picker
- Get margins and paddings working correctly for Picker

### DIFF
--- a/packages/core/src/components/TextField.tsx
+++ b/packages/core/src/components/TextField.tsx
@@ -26,6 +26,7 @@ import {
 } from "@draftbit/types";
 import type { Theme } from "../styles/DefaultTheme";
 import type { IconSlot } from "../interfaces/Icon";
+import { applyStyles, extractStyles } from "../utilities";
 
 const AnimatedText = Animated.createAnimatedComponent(Text);
 
@@ -272,6 +273,10 @@ class TextField extends React.Component<Props, State> {
       backgroundColor = colors.background;
     }
 
+    if (rest.placeholderTextColor) {
+      placeholderColor = rest.placeholderTextColor;
+    }
+
     const { lineHeight, ...subtitle1 } = typography.subtitle1;
 
     inputStyle = {
@@ -398,11 +403,15 @@ class TextField extends React.Component<Props, State> {
       ],
     };
 
-    const inputStyles = [
-      styles.input,
-      inputStyle,
-      type === "solid" ? { marginHorizontal: 12 } : {},
-    ];
+    const { textStyles } = extractStyles(style);
+    const inputStyles = applyStyles(
+      StyleSheet.flatten([
+        styles.input,
+        inputStyle,
+        type === "solid" ? { marginHorizontal: 12 } : {},
+      ]),
+      textStyles
+    );
 
     const {
       backgroundColor: bgColor,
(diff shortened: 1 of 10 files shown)


packages/core/src/utilities.ts
```diff
@@ -32,3 +32,25 @@ export function extractStyles(style: StyleProp<any>) {
 
   return { viewStyles, textStyles };
 }
+
+/**
+ * Merges a style object on top of another style object. In React Native,
+ * keys with undefined values in a style object will still override styles
+ * that appear earlier in a sequence. This avoids that problem.
+ */
+export function applyStyles(
```
dhruska, line 41: This lets us get rid of that `...(something ? { something } : {})` pattern.

As discussed in a previous PR, there doesn't seem to be a better way to do this. These all seem to _not_ work (i.e. they all result in `{ color: undefined }`

- `const mergedStyles = [{ color: "red" }, { color: undefined }]`
- `const mergedStyles = StyleSheet.compose({ color: "red" }, { color: undefined })`
- `const mergedStyles = StyleSheet.flatten([{ color: "red" }, { color: undefined }])`

packages/core/src/components/TextField.tsx
```diff
@@ -427,27 +436,22 @@ class TextField extends React.Component<Props, State> {
...
-                  ...(borderWidth ? { borderWidth } : {}),
-                  ...(borderTopWidth ? { borderTopWidth } : {}),
-                  ...(borderRightWidth ? { borderRightWidth } : {}),
-                  ...(borderBottomWidth ? { borderBottomWidth } : {}),
-                  ...(borderLeftWidth ? { borderLeftWidth } : {}),
-                  ...(borderCol ? { borderColor: borderCol } : {}),
-                }
-              : {},
-          ]}
+          style={applyStyles(StyleSheet.flatten(containerStyle), {
```
peterpme, line 439: do you think we can include the flatten inside applyStyles?

dhruska: Definitely - made this change

packages/core/src/utilities.ts
peterpme: do you mind putting this as a comment in there as an example? i think it'll make more sense

dhruska: 👍 sure thing

packages/core/src/utilities.ts
```diff
@@ -32,3 +32,27 @@ export function extractStyles(style: StyleProp<any>) {
...
+  baseStyles: StyleProp<any>,
+  stylesToApply: StyleProp<any> | undefined
+) {
+  if (!stylesToApply) {
+    return;
+  }
+
+  const flattenedStyles = StyleSheet.flatten(baseStyles);
```
peterpme, line 49: will the base styles always be unflattened? for example, if i run stylesheet.flatten outside of this function?

do you think we should add a check to see if its already flat? maybe this part isn't possible, but i could have sworn styles returns a number, and once flattened, returns an object

adnelson (approved):

packages/core/src/utilities.ts
```diff
@@ -32,3 +32,27 @@ export function extractStyles(style: StyleProp<any>) {
 
   return { viewStyles, textStyles };
 }
+
+/**
+ * Merges a style object on top of another style object. In React Native,
+ * keys with undefined values in a style object will still override styles
+ * that appear earlier in a sequence. This avoids that problem.
+ */
```
dhruska, line 40:
```suggestion
/**
 * Merges a style object on top of another style object. In React Native,
 * keys with undefined values in a style object will still override styles
 * that appear earlier in a sequence. This avoids that problem.
 *
 * This lets us avoid the `...(something ? { something } : {})` pattern.
 * There doesn't seem to be a better way to do this. These all seem to not
 * work (i.e. they all result in `{ color: undefined }`:
 * const mergedStyles = `[{ color: "red" }, { color: undefined }]`
 * const mergedStyles = `StyleSheet.compose({ color: "red" }, { color: undefined })`
 * const mergedStyles = `StyleSheet.flatten([{ color: "red" }, { color: undefined }])`
 */
```

packages/core/src/utilities.ts
dhruska: It seems that, at least on the TS side, `StyleSheet.flatten` expects an array of objects always ((docs)[https://reactnative.dev/docs/stylesheet#flatten]) - so maybe what I can do is enforce the type of `baseStyles` on this function to take an array of objects, so we know that the styles aren't already flattened.

dhruska: 245a9a8

peterpme (approved):

PR #277, opened 2021-07-28
Interactive Star Rating Component

gunjankothari:


packages/core/src/components/StarRating.tsx
```diff
@@ -62,7 +91,7 @@ export const SEED_DATA = {
   name: "Star Rating",
   tag: "StarRating",
   description: "A star rating component",
-  category: COMPONENT_TYPES.basic,
+  category: COMPONENT_TYPES.button,
   props: {
     starSize: createStaticNumberProp({
```
dhruska, line 96: To make this work with Draftbit's state system, you'll also want to remove the "rating" prop below and add (I think):

```suggestion
    fieldName: createFieldNameProp({
      defaultValue: "ratingValue", // this is the name of the variable declared on the screen in Draftbit
      handlerPropName: "onPress", // the change handler prop in this component
      valuePropName: "rating", // the value prop in this component
    }),
    starSize: createStaticNumberProp({
```

packages/core/src/components/StarRating.tsx
```diff
@@ -54,6 +73,16 @@ const styles = StyleSheet.create({
...
   },
+  touchContainer: {
+    display: "flex",
+    flexDirection: "row",
+    position: "absolute",
+    top: 0,
+    right: 0,
+    left: 0,
+    bottom: 0,
+    zIndex: 111,
```
dhruska, line 84: Is it possible to make this a more reasonable value like `1` or `2`?

gunjankothari: Fixed

packages/core/src/components/StarRating.tsx
gunjankothari: updated

dhruska (approved):

PR #279, opened 2021-08-03
Fab icon seed data correction

gunjankothari:
Issue: 
https://community.draftbit.com/c/help/problem-in-fab

This PR should fix the above issue. 
Though I couldn't test this because I don't have access to the builder.

### DIFF
--- a/packages/core/src/components/FAB.tsx
+++ b/packages/core/src/components/FAB.tsx
@@ -110,26 +110,24 @@ const styles = StyleSheet.create({
 
 export default withTheme(FAB);
 
-export const SEED_DATA = [
-  {
-    name: "FAB",
-    tag: "FAB",
-    category: COMPONENT_TYPES.button,
-    description: "A round, mini FAB",
-    layout: {},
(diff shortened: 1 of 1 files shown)


packages/core/src/components/FAB.tsx
```diff
@@ -110,26 +110,24 @@ const styles = StyleSheet.create({
...
-    },
+export const SEED_DATA = {
+  name: "FAB",
+  tag: "FAB",
+  category: COMPONENT_TYPES.button,
+  description: "A round, mini FAB",
+  layout: {},
+  props: {
+    onPress: createActionProp(),
+    icon: createIconProp({
```
dhruska, line 121: I think the problem is that in the above code, the code is looking for a prop called `iconName`, but according to the seed data the prop is (incorrectly) called `icon`

```suggestion
    iconName: createIconProp({
```

gunjankothari: fixed

dhruska (approved):

PR #282, opened 2021-08-18
FAB Icon Absolute Position on Web

gunjankothari:


dhruska:
This seems to work ok for me as-is in Draftbit (in preview mode, not in single screen mode). Because it's a single-screen mode only issue, it may need to be fixed in the internal codebase.

image

PR #285, opened 2021-08-20
P-2514 Slider margin removed from component

gunjankothari:


packages/core/src/components/Slider.tsx
```diff
@@ -122,8 +122,6 @@ const styles = StyleSheet.create({
     height: 40,
     flexDirection: "row",
     alignItems: "center",
-    marginLeft: 12,
-    marginRight: 12,
```
dhruska, line 126: I think the piece left, according to the ticket, would be to make these a part of Seed Data so that they are user-configurable. That would mean changing the `layout` object down on line 139 to be:
```js
layout: {
  marginLeft: 12,
  marginRight: 12,
}
```

gunjankothari: done

dhruska (approved):

PR #289, opened 2021-08-20
P-2487 [Toggle Button] Missing Configs

gunjankothari:


packages/core/src/components/ToggleButton.tsx
```diff
@@ -106,13 +106,18 @@ export const SEED_DATA = {
     disabled: createBoolProp({
       label: "Disabled",
       description: "Whether the button should be disabled",
+      group: GROUPS.basic,
+    }),
+    color: createColorProp({
+      group: GROUPS.basic,
```
dhruska, line 112: Color props created with `createColorProp` should already be in the basic group FYI - though this shouldn't hurt anything

https://github.com/draftbit/react-native-jigsaw/blob/2a6f421fa4e980f332be5b4482b88e90b2034534/packages/types/src/component-types.ts#L236-L246

dhruska (approved):

PR #298, opened 2021-08-31
P-2526 Text Area Component

gunjankothari:


packages/core/src/components/TextField.tsx
```diff
@@ -68,7 +68,7 @@ interface State {
   value?: string;
 }
 
-class TextField extends React.Component<Props, State> {
+class StyledTextField extends React.Component<Props, State> {
```
dhruska, line 71: Can we keep the tag for this as `TextField` under the hood, so that existing TextField components keep working in Draftbit when we upgrade the version of Jigsaw?

```suggestion
class TextField extends React.Component<Props, State> {
```

packages/core/src/components/TextField.tsx
```diff
@@ -595,7 +595,7 @@ class TextField extends React.Component<Props, State> {
   }
 }
 
-export default withTheme(TextField);
+export default withTheme(StyledTextField);
```
dhruska, line 598:
```suggestion
export default withTheme(TextField);
```

packages/core/src/components/TextField.tsx
```diff
@@ -695,8 +695,8 @@ const SEED_DATA_PROPS = {
 const SEED_DATA_TRIGGERS = [Triggers.OnChangeText];
 export const SEED_DATA = [
   {
-    name: "Text Field",
-    tag: "TextField",
+    name: "Styled Text Field",
+    tag: "StyledTextField",
```
dhruska, line 699:
```suggestion
    tag: "TextField",
```

packages/core/src/components/TextField.tsx
```diff
@@ -730,8 +730,8 @@ export const SEED_DATA = [
     layout: {},
   },
   {
-    name: "Text Area",
-    tag: "TextField",
+    name: "Styled Text Area",
+    tag: "StyledTextField",
```
dhruska, line 734:
```suggestion
    tag: "TextField",
```

packages/native/src/components/TextArea.tsx
```diff
@@ -0,0 +1,59 @@
...
+    numberOfLines: {
+      label: "Number Of Lines",
+      description: "Number Of Lines",
+      group: GROUPS.basic,
+      formType: FORM_TYPES.boolean,
+      propType: PROP_TYPES.BOOLEAN,
+      defaultValue: 4,
+      editable: true,
+      required: false,
+    },
```
dhruska, line 55: I think we need the field name prop here, similar to TextInput
```suggestion
    },
    fieldName: {
      ...FIELD_NAME,
      defaultValue: "textInputValue",
      handlerPropName: "onChangeText",
    },
```

packages/core/src/components/TextField.tsx
gunjankothari: Updated

packages/core/src/components/TextField.tsx
gunjankothari: Updated

packages/core/src/components/TextField.tsx
gunjankothari: Updated

packages/core/src/components/TextField.tsx
gunjankothari: Updated

packages/native/src/components/TextArea.tsx
```diff
@@ -0,0 +1,65 @@
...
+    fieldName: {
+      ...FIELD_NAME,
+      defaultValue: "textInputValue",
+      handlerPropName: "onChangeText",
+    },
+  },
+};
```
dhruska, line 63: Rather than creating a new component within draftbit/ui called TextArea, do you think it would be possible to just take this seed data that you created, and put it in a new file at `core/src/mappings/TextArea.js`? That way, you don't need to create a new component, and can re-use the TextInput under the hood.

In other words, a user adding a "Text Area" component will really just be adding a TextInput component, but with `multiline` set to `true` and `numberOfLines` set to `4` by default.

dhruska: Besides moving the seed data to that new file and deleting this file, the one other change you would need to make is making the tag for the component `TextInput` (so that it uses the TextInput component).

packages/core/src/mappings/TextArea.js
```diff
@@ -0,0 +1,46 @@
...
+  PROP_TYPES,
+  FIELD_NAME,
+  TEXT_INPUT_PROPS,
+  Triggers,
+} from "@draftbit/types";
+
+export const SEED_DATA = {
+  name: "Text Area",
+  tag: "TextInput",
+  description: "Render Text Area",
```
dhruska, line 14:
```suggestion
  description: "An input field that allows for multiple lines.",
```

dhruska (approved):

PR #299, opened 2021-09-01
P-2527 Text Field Multiline and Underline Color Props

gunjankothari:


packages/core/src/components/TextField.tsx
```diff
@@ -270,7 +276,7 @@ class TextField extends React.Component<Props, State> {
     } else {
       activeColor = error ? colors.error : colors.primary;
       placeholderColor = borderColor = colors.light;
-      underlineColor = colors.light;
+      underlineColor = underlineColorProp || colors.light;
```
dhruska, line 279:
```suggestion
      underlineColor = underlineColorProp;
```
Could we make this the defaultValue?

packages/core/src/components/TextField.tsx
```diff
@@ -715,6 +722,9 @@ export const SEED_DATA = [
         required: true,
         group: GROUPS.basic,
       },
+      underlineColor: createColorProp({
+        label: "Underline Color",
+      }),
```
dhruska, line 727:
```suggestion
      underlineColor: createColorProp({
        label: "Underline Color",
        defaultValue: "light",
      }),
```

dhruska (approved):

PR #300, opened 2021-09-01
P-2525: Number Input

gunjankothari:


example/src/TextInputExample.js
```diff
@@ -29,6 +30,12 @@ export default function TextInputExample() {
             onChangeText={(text) => setText2(text)}
             initialValue="I'm an initial value!"
           />
+          <NumberInput
+            placeholder="Input with initial value"
```
dhruska, line 34:
```suggestion
            placeholder="Number input with initial value"
```

packages/core/src/mappings/TextInput.js
```diff
@@ -8,186 +8,263 @@ import {
...
-      defaultValue: null,
-      min: 0,
-      step: 1,
-      precision: 0,
-      formType: FORM_TYPES.number,
-      propType: PROP_TYPES.NUMBER,
+    triggers: [Triggers.OnChangeText],
+    props: {
+      ...TEXT_INPUT_PROPS,
+      style: {
```
dhruska, line 35: If these other props, that aren't a part of `TEXT_INPUT_PROPS`, are all shared between both the TextInput and the new Number Input, could you put them in an object above the seed data and then use them in the seed data?

So like:

```js
const NATIVE_INPUT_PROPS = {
  style: {
    whatever
  },
  clearButtonMode: {
    whatever
  }, etc...
};

export const SEED_DATA = [{
  name: "Text Input",
  tag: "TextInput",
  props: {
    ...TEXT_INPUT_PROPS,
    ...NATIVE_INPUT_PROPS,
  }
}, {
  name: "Number Input",
  tag: "NumberInput",
  props: {
    ...TEXT_INPUT_PROPS,
    ...NATIVE_INPUT_PROPS,
  }
];
```

gunjankothari: updated

dhruska (approved):

PR #302, opened 2021-09-06
P-2523 Swiper component

gunjankothari:


example/src/App.tsx
```diff
@@ -139,7 +142,7 @@ function Examples() {
   return (
     <NavigationContainer>
       <Drawer.Navigator
-        initialRouteName="CircleImage"
+        initialRouteName="Swiper"
```
dhruska, line 145:
```suggestion
        initialRouteName="CircleImage"
```

packages/core/src/components/Swiper/SwiperItem.tsx
```diff
@@ -0,0 +1,30 @@
+import { COMPONENT_TYPES } from "@draftbit/types";
+import React from "react";
+import { View, StyleSheet, StyleProp, ViewStyle } from "react-native";
+
+const styles = StyleSheet.create({
+  wrapper: {
+    flex: 1,
+    alignItems: "center",
+    justifyContent: "center",
+  },
```
dhruska, line 10: Should these be user-configurable? If so we should put them in the `layout` object in `SEED_DATA`. If not, disregard

dhruska:
Just one question - this looks good, nice work!

packages/core/src/components/Swiper/SwiperItem.tsx
gunjankothari: Actually, It is not required here, moved it to the example.

dhruska (approved):

PR #306, opened 2021-09-21
P-2378 missing color props for checkbox row component

gunjankothari:


packages/core/src/components/Checkbox/CheckboxRow.tsx
```diff
@@ -184,5 +195,23 @@ export const SEED_DATA = {
       valuePropName: "status",
       handlerPropName: "onPress",
     }),
+    color: createColorProp({
+      group: GROUPS.basic,
```
peterpme, line 199: you can remove group, label, defaultValue from these - the defaults you set are the defaults of the function

peterpme (changes requested):
everything else is good besides the small seed data changes, nice job!

packages/core/src/components/Checkbox/CheckboxRow.tsx
```diff
@@ -184,5 +195,20 @@ export const SEED_DATA = {
...
+      label: "Unselected Color",
+      description: "Color for the button when the checkbox is unchecked",
+      defaultValue: null,
+    }),
+    indeterminateColor: createColorProp({
+      group: GROUPS.basic,
+      label: "Indeterminate Color",
+      description: "Color for the button when the checkbox is indeterminate",
+      defaultValue: null,
```
dhruska, line 211: Peter's comment applies to `uncheckedColor` and `indeterminateColor`, too - `defaultValue` and `group` are the same as the defaults applied in `createColorProp`, so they can be removed here.

gunjankothari: fixed

dhruska (approved):

PR #308, opened 2021-09-23
P-2560 Link Component

gunjankothari:


packages/core/src/components/Text.tsx
```diff
@@ -42,227 +44,267 @@ class Text extends React.Component<Props> {
...
+  theme,
+  title,
+  ...props
+}: any): JSX.Element => {
+  return (
+    <Text
+      hitSlop={8}
+      style={[{ color: theme.colors.primary, alignSelf: "baseline" }, style]}
+      theme={theme}
+      {...props}
```
dhruska, line 58: Does this work right now if you pass an `onPress` prop to it? Does it need to be wrapped in a `Touchable` perhaps? That would also give it the opacity effect when being pressed I imagine.

bluerssen: https://draftbit.slack.com/archives/CTK4DGGS3/p1632251526001000

dhruska: Ah, got it - as long as that works (TIL!) sounds good to me.

dhruska (approved):

PR #311, opened 2021-09-29
swiper component

gunjankothari:


packages/core/src/components/Swiper/Swiper.tsx
```diff
@@ -3,14 +3,16 @@ import {
...
 } from "@draftbit/types";
 import React from "react";
 import { View, StyleSheet, StyleProp, ViewStyle } from "react-native";
 import SwiperComponent from "react-native-web-swiper";
 
 const styles = StyleSheet.create({
   wrapper: {
-    flex: 1,
+    height: 300,
+    width: "100%",
```
dhruska, line 15: I think you can remove these. Since they're in `layout` in seed data, they'll be part of the `style` prop, unless changed by the user. I don't _think_ we want to internally enforce these.

dhruska (approved):

PR #315, opened 2021-09-29
number input props

gunjankothari:


packages/core/src/mappings/TextInput.js
```diff
@@ -222,8 +223,13 @@ export const SEED_DATA = [
...
     triggers: [Triggers.OnChangeText],
     props: {
-      ...TEXT_INPUT_PROPS,
+      ...NUMBER_INPUT_PROPS,
       ...COMMON_NATIVE_INPUT_PROPS,
+      defaultValue: {
+        ...NUMBER_INPUT_PROPS.defaultValue,
+        group: GROUPS.data,
+        label: "Initial Value",
+      },
```
dhruska, line 232: "Initial value" is something we have to turn on within the core Draftbit code for this component, so I don't think you want it as a part of seed data.

Because "initial value" is something we added at Draftbit, and it kind of conflicts with the react-native TextInput `defaultValue`, I think I would just remove the `defaultValue` prop altogether.

gunjankothari: makes sense.

dhruska (approved):

PR #320, opened 2021-10-05
P-2564 Modal component

gunjankothari:


packages/core/src/mappings/Modal.js
```diff
@@ -0,0 +1,38 @@
...
+  name: "Modal",
+  tag: "Modal",
+  description: "A basic Modal Component",
+  category: COMPONENT_TYPES.container,
+  props: {
+    animationType: {
+      group: GROUPS.advanced,
+      name: "animationType",
+      label: "animationType",
+      description: "Animatio Type",
```
dhruska, line 19:
```suggestion
      description: "Animation Type",
```

packages/core/src/mappings/Modal.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    transparent: createBoolProp({
+      group: GROUPS.basic,
+      label: "Transparent",
+      description: "Should be Transparent",
+    }),
+    visible: createBoolProp({
+      group: GROUPS.data,
+      label: "Visible",
+      description: "Should be Visible",
```
dhruska, line 35:
```suggestion
      description: "Determines whether the modal is visible",
```

packages/core/src/mappings/Modal.js
```diff
@@ -0,0 +1,38 @@
...
+      editable: false,
+      required: false,
+      formType: FORM_TYPES.flatArray,
+      propType: PROP_TYPES.STRING,
+      defaultValue: "none",
+    },
+    transparent: createBoolProp({
+      group: GROUPS.basic,
+      label: "Transparent",
+      description: "Should be Transparent",
```
dhruska, line 30: (copied from docs)
```suggestion
      description: "Determines whether the modal will fill the entire view. Setting this to true will render the modal over a transparent background",
```

packages/core/src/mappings/Modal.js
```diff
@@ -0,0 +1,38 @@
...
+} from "@draftbit/types";
+
+export const SEED_DATA = {
+  name: "Modal",
+  tag: "Modal",
+  description: "A basic Modal Component",
+  category: COMPONENT_TYPES.container,
+  props: {
+    animationType: {
+      group: GROUPS.advanced,
```
dhruska, line 16: I already merged this, but I'm not actually seeing this prop anywhere. `Advanced` may be hidden, currently, so I think this needs to be `basic`.

gunjankothari: OK. I will create another PR to make them basic.

PR #327, opened 2021-10-13
textarea numberOfLines formType and PropType num

gunjankothari:


packages/core/src/mappings/TextArea.js
```diff
@@ -31,8 +31,8 @@ export const SEED_DATA = {
       label: "Number Of Lines",
       description: "Number Of Lines",
       group: GROUPS.basic,
-      formType: FORM_TYPES.boolean,
-      propType: PROP_TYPES.BOOLEAN,
+      formType: FORM_TYPES.number,
+      propType: PROP_TYPES.number,
```
dhruska, line 35: Shouldn't this be `PROP_TYPES.NUMBER`?

PR #328, opened 2021-10-13
P-2559 Linear Gradient

gunjankothari:


packages/core/src/components/LinearGradient.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <LinearGradient colors={colors} start={start} end={end} style={viewStyles}>
+      {children}
+    </LinearGradient>
+  );
+};
+
+export default LinearGradientComponent;
+
+export const SEED_DATA = {
```
dhruska, line 48: Could you take this SEED_DATA object and move it over to `packages/core/src/mappings/LinearGradient.ts`? You can replace what's in that file already.

The mappings live in a separate place because we've run into some source map issues in the past where having these objects in here cause issues.

smlarkin: Donezo (also Fugunzo).

dhruska: 🙏

dhruska (approved):

PR #331, opened 2021-10-18
P-2596 Swiper issues

gunjankothari:


packages/core/src/components/Swiper/Swiper.tsx
```diff
@@ -89,6 +102,20 @@ export const SEED_DATA = {
...
     }),
+    prevTitleColor: createColorProp({
+      label: "Previous Title Color",
+    }),
+    nextTitleColor: createColorProp({
+      label: "Next Title Color",
+    }),
+    dotColor: createColorProp({
+      label: "Dot Color",
+      defaultValue: "#86939e",
```
dhruska, line 113: I don't think hex default values will work. It needs to be one of the theme colors e.g. `primary`, `secondary`, `error` etc. But run the script against staging, maybe I'm wrong

PR #343, opened 2021-11-08
Make checkbox take a boolean instead of string

dhruska:


packages/core/src/mappings/Checkbox.js
```diff
@@ -17,7 +17,7 @@ export const SEED_DATA = {
   props: {
     fieldName: createFieldNameProp({
       defaultValue: "checkboxValue",
-      valuePropName: "status",
+      valuePropName: "value",
```
dhruska, line 20:
```suggestion
      valuePropName: "status",
```

packages/core/src/components/Switch.tsx
```diff
@@ -116,13 +117,19 @@ function Row({
   theme,
   ...rest
 }: Props & RowProps) {
-  const [checked, setChecked] = React.useState(value);
+  const [checked, setChecked] = React.useState(value || defaultValue);
```
adnelson, line 120:
```suggestion
  const [checked, setChecked] = React.useState(value != null ? value : defaultValue);
```

Just thinking that since value is a bool, as written this would ignore `false`

dhruska: Ooh, good point!

adnelson (approved):

PR #346, opened 2021-11-10
P-2587 Data Table Component

gunjankothari:
Snack: https://snack.expo.dev/@gunjankothari/jigsaw---table-component


packages/core/src/components/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import { View, Text } from "react-native";
+
+const TablePaginator = () => (
```
bluerssen, line 4: Is this file needed? I dont get what it does.

packages/core/src/components/Table/TableHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+);
+
+const styles = StyleSheet.create({
+  wrapper: {
+    display: "flex",
+    flexDirection: "row",
+  },
+});
+
+export default withTheme(TableHeader);
```
bluerssen, line 28: Lets remove theme and theme objects from these and instead make sure we can just style them with props, and set reasonable defaults on those props.

example/src/TableExample.js
```diff
@@ -0,0 +1,207 @@
...
+  },
+  cell: {
+    paddingTop: 16,
+    paddingBottom: 16,
+    paddingLeft: 8,
+    paddingRight: 8,
+  },
+});
+
+export default withTheme(TableExample);
```
bluerssen, line 207: Lets remove theme and theme objects from these and instead make sure we can just style them with props, and set reasonable defaults on those props.

packages/core/src/utilities.ts
```diff
@@ -1,5 +1,13 @@
 import { StyleSheet, StyleProp, TextStyle } from "react-native";
-import { isString, isNumber, pick, pickBy, identity } from "lodash";
+import {
+  isString,
+  isNumber,
+  pick,
+  pickBy,
+  identity,
+  omitBy,
```
bluerssen, line 8: In the interest of getting rid of things that might be causing the import errors, do we need these new ones?

packages/core/src/components/Table/TablePaginator.tsx
gunjankothari: I had created this file initially thinking we might support pagination but we decided to go without pagination that time. I can remove this file.

packages/core/src/utilities.ts
gunjankothari: these are just lodash methods. 
Purpose of this is to remove properties which are not defined. 

Do you think this can be causing problem?

packages/core/src/components/Table/TableHeader.tsx
gunjankothari: removed

example/src/TableExample.js
gunjankothari: removed

packages/core/src/mappings/Table.ts
```diff
@@ -0,0 +1,145 @@
...
+      StylesPanelSections.Background,
+      StylesPanelSections.Borders,
+      StylesPanelSections.Size,
+      StylesPanelSections.MarginsAndPaddings,
+      StylesPanelSections.Position,
+      StylesPanelSections.Effects,
+    ],
+    layout: {
+      fontSize: 14,
+      fontWeight: "500",
```
bluerssen, line 116: @adnelson

YoussefHenna:
Closed in favor of https://github.com/draftbit/react-native-jigsaw/pull/598

PR #347, opened 2021-11-11
Remove initialValue prop

dhruska:
It's been deprecated in favor of `defaultValue` and is no longer in use by Draftbit


packages/core/src/components/Checkbox/CheckboxGroup.tsx
```diff
@@ -76,7 +76,7 @@ export const SEED_DATA = {
     fieldName: createFieldNameProp({
       defaultValue: "checkboxGroupValue",
       handlerPropName: "onValueChange",
-      valuePropName: "checkboxGroupValue",
```
dhruska, line 79: This was just a bug, the underlying component actually expects a `value` prop already

example/src/CheckboxExample.js
```diff
@@ -34,6 +36,14 @@ const CheckboxExample = ({ theme }) => {
   };
 
   const handlePress = (value) => setChecked(value);
+  const handleAirtablePress = (value) => {
+    console.log("handling airtable press", value);
```
dhruska, line 40:
```suggestion
```

example/src/CheckboxExample.js
```diff
@@ -34,6 +36,13 @@ const CheckboxExample = ({ theme }) => {
   };
 
   const handlePress = (value) => setChecked(value);
+  const handleAirtablePress = (value) => {
```
dhruska, line 39:
```suggestion
  // An example to simulate how Airtable returns boolean values: `true` or `undefined`
  const handleAirtablePress = (value) => {
```

example/src/CheckboxExample.js
```diff
@@ -21,6 +22,7 @@ const SingleCheckboxWrapper = ({ label, children }) => (
 
 const CheckboxExample = ({ theme }) => {
   const [checked, setChecked] = React.useState(true);
+  const [airtableChecked, setAirtableChecked] = React.useState(false);
```
dhruska, line 25:
```suggestion
  const [airtableChecked, setAirtableChecked] = React.useState(undefined);
```

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -42,12 +39,12 @@ const RadioButtonGroup: React.FC<RadioButtonGroupProps> = ({
     }
   }, [defaultValue]);
 
-  const previousInitialValue = usePrevious(initialValue);
-  React.useEffect(() => {
-    if (initialValue !== previousInitialValue) {
-      onValueChange(initialValue);
+  const handleValueChange = (newValue: string) => {
+    setInternalValue(newValue);
+    if (onValueChange) {
```
adnelson, line 44: `onValueChange` is defaulted to a no-op function; I don't think we need to worry about it being undefined here right?

dhruska: Yeah, you're right. That check should be fine to remove

packages/core/src/components/TextField.tsx
```diff
@@ -97,9 +96,6 @@ class TextField extends React.Component<Props, State> {
   componentDidMount() {
     if (this.props.defaultValue) {
       this._handleChangeText(this.props.defaultValue);
```
adnelson, line 98: I didn't see functionality to react to changes to `defaultValue` here -- are we doing that? Looks like this is just firing on mount?

packages/core/src/components/TextInput.tsx
```diff
@@ -1,25 +0,0 @@
-import React from "react";
```
adnelson, line 1: why are we deleting this file?

packages/core/src/components/TextField.tsx
dhruska: If we're going to update the react-query centric logic to use `onChange` and `value`, I don't think there's a need to react to changes in `defaultValue` (which would mirror the behavior of the react native TextInput component)

packages/core/src/components/TextInput.tsx
dhruska: Since we're removing `initialValue`, there's no longer a need for a wrapper component. We can just use the react-native TextInput

packages/core/src/components/TextField.tsx
dhruska: We could add it though, I suppose? I don't think it would hurt anything

adnelson (dismissed):

adnelson (approved):

PR #353, opened 2021-11-16
Move all seed data into mapping files and convert to TS

dhruska:
- Convert existing mappings to ts
- move over remaining seed data


packages/core/src/mappings/RadioButtonRow.ts
```diff
@@ -0,0 +1,46 @@
...
+    }),
+    unselectedColor: createColorProp({
+      group: GROUPS.basic,
+      label: "Unselected Color",
+      description: "Unselected Color for the button",
+      defaultValue: "primary",
+    }),
+  },
+};
+// TODO comment testing
```
dhruska, line 46:
```suggestion
```

peterpme (dismissed):

PR #358, opened 2021-11-18
P-2634 Text Area Placeholder

gunjankothari:


packages/core/src/mappings/TextArea.ts
```diff
@@ -17,6 +17,11 @@ export const SEED_DATA = {
   triggers: [Triggers.OnChangeText],
   props: {
     ...TEXT_INPUT_PROPS,
+    placeholder: {
+      ...TEXT_INPUT_PROPS.placeholder,
+      defaultValue:
```
dhruska, line 22: Maybe we should just change this in the source of the TEXT_INPUT_PROPS and NUMBER_INPUT_PROPS? https://github.com/draftbit/react-native-jigsaw/blob/b029f74dc5328c12dc2c23dfcb6a219df70dd803/packages/types/src/component-types.ts#L493

dhruska (approved):

PR #372, opened 2021-12-09
correctly pass in either defaultValue or value

adnelson:
This might be a little overcomplicated given that we don't need to watch `defaultValue` for changes, but at least it works:

* Supports integers
* Supports floats (typing in decimals)
* Shows correct `value` and updates it when it changes
* Supports `value` and `defaultValue` following the same rules as react-native TextInput


dhruska (approved):
Thanks for the examples + running through this on video. And for tackling this one 💀 

LGTM!

dhruska (approved):

PR #376, opened 2021-12-20
P-2683 Keyboard Aware Scroll View

gunjankothari:


smlarkin:
keyboard avoiding view added directly in draftbit/builder with the [Add keyboardAwareScrollView](https://github.com/draftbit/draftbit/pull/5221) PR so as not to make it a dependency of Jigsaw.

PR #380, opened 2021-12-23
fix-stepper-min-max-value

karimandrusevych:


dhruska:
Looks like this functionality was also added in #381, so closing this one

PR #384, opened 2022-01-06
Add support for onCheck/onUncheck triggers in Checkbox, remove ToggleButton

dhruska:
- Removes ToggleButton because its intended behavior is already handled by Checkbox
- Adds `onCheck` and `onUncheck` props to Checkbox so actions can easily be attached to it that should only be fired in check or uncheck situations
- Enables `onPressIcon` prop on Card (though card is currently hidden)


packages/core/src/mappings/Card.ts
```diff
@@ -16,7 +16,7 @@ export const SEED_DATA = {
   tag: "Card",
   description: "A card you can customize however you'd like",
   category: COMPONENT_TYPES.card,
-  triggers: [Triggers.OnPress], //TODO: when we support multiple triggers, add Triggers.OnPressIcon
+  triggers: [Triggers.OnPress, Triggers.OnPressIcon],
```
dhruska, line 19: this component is currently hidden but I made this work anyways in case we unhide it down the road

peterpme (approved):

PR #385, opened 2022-01-07
Fix FAB icon sizing

smlarkin:
Per [P-2684](https://linear.app/draftbit/issue/P-2684/fab-icon-size-prop-not-working), the `FAB` component sizing prop did not grow the inner `Icon` to match the outer/background `View`.

- Updated `Icon` size prop
- Updated `ActivityIndicator` style & size props
- Updated the example with 3 sizes each doubling -- 25px, 50px (default prop), 100px -- to show how it scales.


dhruska (approved):

peterpme (approved):
LGTM!

PR #389, opened 2022-01-13
Move LinearGradient to native directory

smlarkin:
Fixes issues with web CRA compatibility.


packages/native/src/components/LinearGradient.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { StyleProp, ViewStyle } from "react-native";
 import { LinearGradient } from "expo-linear-gradient";
-import { extractStyles } from "../utilities";
+// import { extractStyles } from "../../../core/src/utilities";
```
peterpme, line 4: if you need this file, just copy it over into utilities to make it easy for yourself. It's ok

packages/ui/src/index.tsx
```diff
@@ -38,7 +38,6 @@ export {
   ActionSheetCancel,
   Swiper,
   SwiperItem,
-  LinearGradient,
```
peterpme, line 41: this still needs to exist, just under native instead of core

example/src/LinearGradientExample.js
```diff
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { Text } from "react-native";
-import { ScreenContainer, withTheme, LinearGradient } from "@draftbit/ui";
+import { ScreenContainer, withTheme } from "@draftbit/ui";
```
peterpme, line 3: needs to exist from UI, just exported from native instead of core

packages/native/src/components/LinearGradient.d.ts
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
```
peterpme, line 1: did this generate by itself or did you create it by hand?

smlarkin: @peter I made it.

packages/native/src/components/LinearGradient.tsx
smlarkin: @peter I don't know if I need this file.  You did not use this option for other native files.

packages/ui/src/index.tsx
smlarkin: @peterpme Gotcha.

example/src/LinearGradientExample.js
smlarkin: @peterpme Gotcha.

packages/native/src/components/LinearGradient.d.ts
smlarkin: @peterpme please advise (should I keep it?)

packages/native/src/components/LinearGradient.tsx
smlarkin: @peterpme please advise ... should I use `extractStyles` here too ?

packages/native/src/components/LinearGradient.d.ts
peterpme: I think so!

packages/native/src/components/LinearGradient.tsx
peterpme: If you don't need to extract the view styles from linear gradient then you can pass down the props regularly set style={style}

We extract the view styles on components when there's children components that need separate props.

smlarkin: @peterpme Damn it!  I just added it.

smlarkin: @peter I'll let you make the final call on the review ... keep / leave ...

packages/native/src/components/LinearGradient.tsx
```diff
@@ -24,14 +24,14 @@ const LinearGradientComponent = ({
   startY = 0,
   endX = 100,
   endY = 100,
-  style,
+  style: propStyle,
 }: LinearGradientComponentProps) => {
-  const { viewStyles } = extractStyles(style);
+  const { viewStyles: style } = extractStyles(propStyle);
```
peterpme, line 29: in this specific case you don't need to use extractStyles since there is no other styles being applied anywhere else.

Cases where extractStyles is useful:

- FormRow, where we must separate viewStyle from textStyle. We separate the two because they're applied to two separate elements:
https://github.com/draftbit/react-native-jigsaw/blob/a421be23d58d60b71257a16eb8db11d13a4713bb/packages/core/src/components/FormRow.tsx#L25

packages/native/src/components/LinearGradient.tsx
```diff
@@ -24,14 +24,14 @@ const LinearGradientComponent = ({
...
   const colors = [color1, color2, color3].filter((color) => color) as string[];
   const start = { x: startX / 100, y: startY / 100 };
   const end = { x: endX / 100, y: endY / 100 };
   return (
-    <LinearGradient colors={colors} start={start} end={end} style={viewStyles}>
+    <LinearGradient {...{ colors, start, end, style }}>
```
peterpme, line 34:
```suggestion
    <LinearGradient colors={colors} start={start} end={end} style={style}>
```

packages/native/src/utilities.ts
```diff
@@ -0,0 +1,65 @@
+import { StyleSheet, StyleProp, TextStyle } from "react-native";
```
peterpme, line 1: since you don't need to use extractStyles in that other component, you can delete this file lol

packages/native/src/components/LinearGradient.tsx
smlarkin: @peter unless there is a perf diff, this is just style now.

smlarkin: @peter see latest commit

adnelson (approved):
😎

PR #390, opened 2022-01-14
Update mappings and default props on LinearGradient

smlarkin:


peterpme (approved):

packages/core/src/mappings/LinearGradient.ts
```diff
@@ -9,71 +9,75 @@ export const SEED_DATA = {
...
       formType: FORM_TYPES.number,
       propType: PROP_TYPES.NUMBER,
     },
+    color3: createColorProp({
+      label: "Color 3",
+      defaultValue: undefined,
+    }),
+    color2: createColorProp({
+      label: "Color 2",
+      defaultValue: "rgba(251, 252, 253, 1)",
```
dhruska, line 76: Does this work ok inside the builder experience after you run the script? I'm pretty sure we typically use theme colors for these, e.g. `"primary"`, which will translate to `theme.colors.primary`

dhruska (approved):

PR #394, opened 2022-01-24
Allow radioButton(Row/Group) zero & number values

smlarkin:
Per [P-2714](https://linear.app/draftbit/issue/P-2714/cant-select-radio-button-when-value-is-set-to-0-zero), assigning a value of 0 (not a sting "0") is not supported with any of the RadioButton components.

Although they are typed for String(s), when strictly using JS, they will allow a user to initialize / use number props and work -- even though real forms do not use numbers as values.  However, 0 being a falsely value is not recognized and is rejected when you enter it in as a prop.

Hence, the solution to parse all numbers to strings and so that they if they are provided in a JS (non-TS) context, they will work as expected -- including the number 0.

### DIFF
--- a/example/src/RadioButtonExample.js
+++ b/example/src/RadioButtonExample.js
@@ -54,9 +54,9 @@ const RadioButtonGroupExample = ({ theme }) => {
           value={selected}
         >
           <RadioButtonRow label="Zero" value={0} />
-          <RadioButtonRow label="1st" value={1} />
+          <RadioButtonRow label="1st" value={"1"} />
           <RadioButtonRow label="2nd" value={2} />
-          <RadioButtonRow label="3rd" value={3} />
+          <RadioButtonRow label="3rd" value={"3"} />
         </RadioButtonGroup>
       </Section>
 
--- a/example/src/RadioButtonExample.js
+++ b/example/src/RadioButtonExample.js
@@ -21,7 +21,7 @@ const SingleRadioButtonWrapper = ({ label, children }) => (
 );
 
 const RadioButtonGroupExample = ({ theme }) => {
-  const [selected, onSelect] = React.useState("1");
+  const [selected, onSelect] = React.useState("0");
   const [selected2, onSelect2] = React.useState("1");
   const handleSelect = (value) => onSelect(value);
   return (
--- a/example/src/RadioButtonExample.js
+++ b/example/src/RadioButtonExample.js
@@ -53,9 +53,10 @@ const RadioButtonGroupExample = ({ theme }) => {
           onValueChange={handleSelect}
           value={selected}
         >
-          <RadioButtonRow label="First" value="1" />
-          <RadioButtonRow label="Second" value="2" />
-          <RadioButtonRow label="Third" value="3" />
+          <RadioButtonRow label="Zero" value={0} />
+          <RadioButtonRow label="1st" value={1} />
+          <RadioButtonRow label="2nd" value={2} />
+          <RadioButtonRow label="3rd" value={3} />
         </RadioButtonGroup>
       </Section>
 
--- a/packages/core/src/components/RadioButton/RadioButton.tsx
+++ b/packages/core/src/components/RadioButton/RadioButton.tsx
@@ -27,7 +27,7 @@ const RadioButton: React.FC<RadioButtonProps> = ({
   value,
   selected,
   unselectedColor,
-  onPress = () => {},
+  onPress,
   size = Config.radioButtonSize,
   selectedIcon = "MaterialIcons/radio-button-checked",
   unselectedIcon = "MaterialIcons/radio-button-unchecked",
@@ -36,16 +36,16 @@ const RadioButton: React.FC<RadioButtonProps> = ({
 }) => {
   const { value: contextValue, onValueChange } = useRadioButtonGroupContext();
 
-  const handlePress = () => {
-    onPress && onPress();
+  // @ts-expect-error
+  const valueExists = value === 0 || value;
 
-    if (onValueChange && value) {
-      onValueChange(value);
-    }
+  const handlePress = () => {
+    onPress?.();
+    valueExists && onValueChange(value);
   };
 
   const isSelected =
-    selected || (contextValue != null && contextValue === value);
+    selected || (contextValue !== null && contextValue === value);
 
   return (
     <IconButton
--- a/packages/core/src/components/RadioButton/RadioButton.tsx
+++ b/packages/core/src/components/RadioButton/RadioButton.tsx
@@ -12,7 +12,7 @@ export type RadioButtonProps = {
   selected?: boolean;
   disabled?: boolean;
   color?: string;
-  value?: string | number;
+  value: string | number;
   unselectedColor?: string;
   onPress?: (value: string) => void;
   style?: StyleProp<ViewStyle>;
@@ -25,7 +25,7 @@ const RadioButton: React.FC<RadioButtonProps> = ({
   Icon,
   disabled = false,
   color,
-  value,
+  value = "",
   selected,
   unselectedColor,
   onPress,
@@ -39,15 +39,11 @@ const RadioButton: React.FC<RadioButtonProps> = ({
 
   const realValue = getValueForRadioButton(value);
   const realContextValue = getValueForRadioButton(contextValue);
-  const isSelected =
-    selected ??
-    (realContextValue && realValue && realContextValue === realValue);
+  const isSelected = selected ?? realContextValue === realValue;
 
   const handlePress = () => {
-    if (realValue) {
-      onPress?.(realValue);
-      onValueChange?.(realValue);
-    }
+    onPress?.(realValue);
+    onValueChange?.(realValue);
   };
 
   return (
--- a/packages/core/src/components/RadioButton/RadioButton.tsx
+++ b/packages/core/src/components/RadioButton/RadioButton.tsx
@@ -12,9 +12,9 @@ export type RadioButtonProps = {
   selected?: boolean;
   disabled?: boolean;
   color?: string;
-  value: string | number;
+  value?: string | number;
   unselectedColor?: string;
-  onPress?: (value: string) => void;
+  onPress?: (value?: string) => void;
   style?: StyleProp<ViewStyle>;
   size?: number;
   selectedIcon?: string;
--- a/packages/core/src/components/RadioButton/RadioButton.tsx
+++ b/packages/core/src/components/RadioButton/RadioButton.tsx
@@ -3,6 +3,7 @@ import { StyleProp, ViewStyle } from "react-native";
 
 import Config from "../Config";
 import IconButton from "../IconButton";
+import { getRealValue } from "../../utilities";
 import { useRadioButtonGroupContext } from "./context";
 
 import type { IconSlot } from "../../interfaces/Icon";
@@ -13,7 +14,7 @@ export type RadioButtonProps = {
   color?: string;
   value?: string;
   unselectedColor?: string;
-  onPress?: () => void;
+  onPress?: (value: string) => void;
   style?: StyleProp<ViewStyle>;
   size?: number;
   selectedIcon?: string;
@@ -36,16 +37,22 @@ const RadioButton: React.FC<RadioButtonProps> = ({
 }) => {
   const { value: contextValue, onValueChange } = useRadioButtonGroupContext();
 
-  // @ts-expect-error
-  const valueExists = value === 0 || value;
+  const realValue = getRealValue(value);
+  const realContextValue = getRealValue(contextValue);
+  const isSelected =
+    selected ??
+    (realContextValue && realValue && realContextValue === realValue);
+
+  console.log({ realValue, realContextValue, isSelected });
 
   const handlePress = () => {
-    onPress?.();
-    valueExists && onValueChange(value);
-  };
+    console.log("RadioButton:realValue", realValue);
 
-  const isSelected =
-    selected || (contextValue !== null && contextValue === value);
+    if (realValue) {
+      onPress?.(realValue);
+      onValueChange?.(realValue);
+    }
+  };
 
   return (
     <IconButton
--- a/packages/core/src/components/RadioButton/RadioButton.tsx
+++ b/packages/core/src/components/RadioButton/RadioButton.tsx
@@ -3,7 +3,7 @@ import { StyleProp, ViewStyle } from "react-native";
 
 import Config from "../Config";
 import IconButton from "../IconButton";
-import { getRealValue } from "../../utilities";
+import { getValueForRadioButton } from "../../utilities";
 import { useRadioButtonGroupContext } from "./context";
 
 import type { IconSlot } from "../../interfaces/Icon";
@@ -12,7 +12,7 @@ export type RadioButtonProps = {
   selected?: boolean;
   disabled?: boolean;
   color?: string;
-  value?: string;
+  value?: string | number;
   unselectedColor?: string;
   onPress?: (value: string) => void;
   style?: StyleProp<ViewStyle>;
@@ -37,8 +37,8 @@ const RadioButton: React.FC<RadioButtonProps> = ({
 }) => {
   const { value: contextValue, onValueChange } = useRadioButtonGroupContext();
 
-  const realValue = getRealValue(value);
-  const realContextValue = getRealValue(contextValue);
+  const realValue = getValueForRadioButton(value);
+  const realContextValue = getValueForRadioButton(contextValue);
   const isSelected =
     selected ??
     (realContextValue && realValue && realContextValue === realValue);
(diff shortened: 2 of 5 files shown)


packages/core/src/utilities.ts
```diff
@@ -63,3 +63,16 @@ export function applyStyles(
 
   return flattenedStyles;
 }
+
+export function getRealValue(value: any) {
```
peterpme, line 67:
```suggestion
export function getRealValue(value: string | number) {
```

i think ideally you want a string or number right?

packages/core/src/utilities.ts
```diff
@@ -63,3 +63,16 @@ export function applyStyles(
...
+  // console.log("getRealValue typeof", typeof value);
+  // console.log("getRealValue value", value);
+  switch (typeof value) {
```
peterpme, line 70: i would personally use lodash' `isString` or `isNumber` which does some fancier stuff under the hood to absolutely make sure:

you can install:
lodash.isstring
lodash.isnumber

we use these in a couple of places

and then 
```
if (!isString(x) && !Number(x)) { throw ... }
```

packages/core/src/utilities.ts
smlarkin: @peterpme done.

packages/core/src/utilities.ts
smlarkin: @peterpme done.

Also, you might want to consider this lib over lodash for the future: https://anguscroll.com/just/

packages/core/src/utilities.ts
```diff
@@ -63,3 +63,16 @@ export function applyStyles(
 
   return flattenedStyles;
 }
+
+export function getRealValue(value: any) {
+  // console.log("getRealValue typeof", typeof value);
+  // console.log("getRealValue value", value);
```
dhruska, line 69: I think you can remove these comments?

packages/core/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -22,7 +22,7 @@ export enum Direction {
 
 export interface RadioButtonRowProps extends Omit<RadioButtonProps, "onPress"> {
   label: string | React.ReactNode;
-  value: string; // A string that this radio button row represents when selected
+  value: string; // A string (or number that will be parsed String(number)) that this radio button row represents when selected
```
dhruska, line 25: If a number is passed in here (like in the example screen), shouldn't this be `value: string | number;`?

packages/core/src/utilities.ts
```diff
@@ -63,3 +64,10 @@ export function applyStyles(
 
   return flattenedStyles;
 }
+
+export function getRealValue(value: string | number | undefined) {
+  if (isString(value)) return value;
+  else if (isNumber(value)) return String(value);
+  else if (!value) return undefined;
```
peterpme, line 71: whats the purpose of undefined? is that bc nothing is selectedin the radio button, which makes it undefined?

does null work here too?

packages/core/src/utilities.ts
```diff
@@ -63,3 +64,10 @@ export function applyStyles(
 
   return flattenedStyles;
 }
+
+export function getRealValue(value: string | number | undefined) {
+  if (isString(value)) return value;
```
peterpme, line 69: just some formatting / style:

```
if (isString(value)) {
  return value
}

if (isNumber(value)) {
  return String(value)
}
```

not sure why the eslint stuff isn't picking up here in jigsaw (perhaps its missing) but in the builder we don't allow 1 line if statements

wrt else if: since the previous if block is already returning, you can avoid the `else if`

packages/core/src/utilities.ts
```diff
@@ -63,3 +64,10 @@ export function applyStyles(
 
   return flattenedStyles;
 }
+
+export function getRealValue(value: string | number | undefined) {
```
peterpme, line 68: since this is only used for radio buttons (at least for now) do you mind calling this `getValueForRadioButton`? or `getValueForRadio)`?

packages/core/src/utilities.ts
smlarkin: @peter @dhruska You can also pass in an undefined and get back an undefined, which it now accepts as an optional type arg.  I used Peter's notes and added an error if the arg is not type undefined, string, or number :)

packages/core/src/components/RadioButton/RadioButtonRow.tsx
smlarkin: @donald done.

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -17,31 +18,39 @@ const { Provider } = radioButtonGroupContext;
...
 
-  const handleValueChange = (newValue: string) => {
-    setInternalValue(newValue);
-    onValueChange(newValue);
+  const handleValueChange = (newValue: any) => {
+    const realNewValue = getRealValue(newValue);
+
+    if (realNewValue) {
+      setInternalValue(newValue);
+      onValueChange?.(newValue);
```
peterpme, line 52: any reason you removed that default function to replace it with this?

packages/core/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -79,9 +79,17 @@ const RadioButtonRow: React.FC<RadioButtonRowProps & IconSlot> = ({
...
+  const realContextValue = getRealValue(contextValue);
+  const isSelected =
+    selected ??
+    (realContextValue && realValue && realContextValue === realValue);
+
   const handlePress = () => {
-    onPress(value);
-    onValueChange && onValueChange(value);
+    if (realValue) {
+      onPress?.(realValue);
```
peterpme, line 90: any reason you removed the default function to press with optional?

packages/core/src/utilities.ts
smlarkin: @peter I will keep it consistent with people that style if + if else + else

packages/core/src/utilities.ts
smlarkin: undefines were being used as conditionals for value changes

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
smlarkin: @peter It's just a safety to be sure the value is not undefined.

packages/core/src/components/RadioButton/RadioButtonRow.tsx
smlarkin: @peter It's just a safety to be sure the value is not undefined.

packages/core/src/utilities.ts
smlarkin: @peter done.

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
smlarkin: @peter Good to go.

packages/core/src/components/RadioButton/RadioButtonRow.tsx
smlarkin: @peter Good to go.

packages/core/src/utilities.ts
smlarkin: @peter. Done.

packages/core/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -112,12 +116,10 @@ const RadioButtonRow: React.FC<RadioButtonRowProps & IconSlot> = ({
       >
         <RadioButton
           Icon={Icon}
-          selected={
-            selected || (contextValue != null && contextValue === value)
-          }
+          selected={isSelected ? true : false}
```
peterpme, line 119:
```suggestion
          selected={isSelected}
```

packages/core/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -46,24 +46,24 @@ const getRadioButtonAlignment = (
...
 
 const renderLabel = (
-  value: string | React.ReactNode,
+  label: string | React.ReactNode,
   labelStyle: StyleProp<TextStyle>,
   textStyle: StyleProp<TextStyle>
 ) => {
-  if (typeof value === "string") {
-    return <Text style={[labelStyle, textStyle]}>{value}</Text>;
+  if (typeof label === "string") {
```
peterpme, line 53: are you fixing a bug here? this might be here on purpose

packages/core/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -46,24 +46,24 @@ const getRadioButtonAlignment = (
...
 
 const RadioButtonRow: React.FC<RadioButtonRowProps & IconSlot> = ({
   Icon,
   label,
-  value,
+  value = "",
   color,
   unselectedColor,
-  onPress = () => {},
+  onPress,
```
peterpme, line 66: why remove the default function?

packages/core/src/components/RadioButton/RadioButtonRow.tsx
```diff
@@ -112,12 +116,10 @@ const RadioButtonRow: React.FC<RadioButtonRowProps & IconSlot> = ({
...
         <RadioButton
           Icon={Icon}
-          selected={
-            selected || (contextValue != null && contextValue === value)
-          }
+          selected={isSelected ? true : false}
+          value={realValue}
           color={color}
           unselectedColor={unselectedColor}
-          onPress={handlePress}
```
peterpme, line 120: where is this onPress being called if its not here?

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -60,7 +61,7 @@ const RadioButtonGroup: React.FC<RadioButtonGroupProps> = ({
     <View style={[{ minHeight: 40 }, style]} {...rest}>
       <Provider
         value={{
-          value: internalValue || "",
+          value: internalValue ?? "",
```
peterpme, line 64: any reason you're changing this to `??`?

example/src/RadioButtonExample.js
```diff
@@ -53,9 +53,10 @@ const RadioButtonGroupExample = ({ theme }) => {
           onValueChange={handleSelect}
           value={selected}
         >
-          <RadioButtonRow label="First" value="1" />
-          <RadioButtonRow label="Second" value="2" />
-          <RadioButtonRow label="Third" value="3" />
+          <RadioButtonRow label="Zero" value={0} />
+          <RadioButtonRow label="1st" value={1} />
+          <RadioButtonRow label="2nd" value={2} />
+          <RadioButtonRow label="3rd" value={3} />
```
peterpme, line 59: since you're using string | number, can you set one of these examples to use a string?

packages/core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -11,9 +12,9 @@ export type RadioButtonProps = {
   selected?: boolean;
   disabled?: boolean;
   color?: string;
-  value?: string;
+  value: string | number;
```
peterpme, line 15: any reason you've removed the optional part of this? i don't think it will always include the value prop

packages/core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -24,10 +25,10 @@ const RadioButton: React.FC<RadioButtonProps> = ({
   Icon,
   disabled = false,
   color,
-  value,
+  value = "",
   selected,
   unselectedColor,
-  onPress = () => {},
+  onPress,
```
peterpme, line 31: any reason you've removed the default function? I think it was there for a reason, unless thats part of the bug fix

packages/core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -36,17 +37,15 @@ const RadioButton: React.FC<RadioButtonProps> = ({
 }) => {
   const { value: contextValue, onValueChange } = useRadioButtonGroupContext();
 
-  const handlePress = () => {
-    onPress && onPress();
+  const realValue = getValueForRadioButton(value);
+  const realContextValue = getValueForRadioButton(contextValue);
+  const isSelected = selected ?? realContextValue === realValue;
```
peterpme, line 42: this is a different operation than what used to be here - is that on purpose?

```
const isSelected = selected || (contextValue != null && contextValue === value);

// const isSelected = selected || (contextValue != null && contextValue === value);

```

```
const isSelected = selected ?? realContextValue === realValue;

// const isSelected = selected !== null && selected !== void 0 ? selected : realContextValue === realValue;
```

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -16,32 +17,32 @@ const { Provider } = radioButtonGroupContext;
...
-  const [internalValue, setInternalValue] = React.useState<string | undefined>(
-    value || defaultValue
-  );
+  const [internalValue, setInternalValue] = React.useState<string>("");
 
   React.useEffect(() => {
-    if (value != null) {
-      setInternalValue(value);
-    }
+    const realValue = getValueForRadioButton(value);
```
peterpme, line 30: any reason you got rid of that `if (value != null)` here?

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
```diff
@@ -16,32 +17,32 @@ const { Provider } = radioButtonGroupContext;
...
+
+    setInternalValue(realValue);
   }, [value]);
 
   React.useEffect(() => {
-    if (defaultValue != null) {
-      setInternalValue(defaultValue);
-    }
+    const realDefaultValue = getValueForRadioButton(defaultValue);
```
peterpme, line 36: same here

packages/core/src/components/RadioButton/RadioButtonRow.tsx
smlarkin: @peter Not needed.  if a real one is passed in it is called for real.  See line 87.

packages/core/src/components/RadioButton/RadioButtonRow.tsx
smlarkin: @Peter only changed the name a variable for troubleshooting ... changed it back for your :)

packages/core/src/components/RadioButton/RadioButtonRow.tsx
smlarkin: @peterpme  It's at the topmost level of this component tree with the `Touchable` line 95 ... I had found some issues conflicting having both the Row control it and the Button control the component.

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
smlarkin: @peter I like it.  But it is changed back because you asked.

example/src/RadioButtonExample.js
smlarkin: @peterpme Done.

packages/core/src/components/RadioButton/RadioButton.tsx
smlarkin: @peter Unneeded. If one exists it will run.  See line 45.

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
smlarkin: @peter Same thing since the real value would have thrown if it is undefined or null

packages/core/src/components/RadioButton/RadioButtonGroup.tsx
smlarkin: @peterpme Same thing since the real value would have thrown if it is undefined or null

packages/core/src/components/RadioButton/RadioButton.tsx
smlarkin: @peterpme this is the original code...

`  const handlePress = () => {
    onPress && onPress();

    if (onValueChange && value) {
      onValueChange(value);
    }
  };

  const isSelected =
    selected || (contextValue != null && contextValue === value);`
    
... firstly, onPress will be called if it is passed in / exists now .. that is what this does: `onPress?.(realValue);`

Before you were not passing the value into onPress ... that looked like a mistake to me.  It looked like if you provided an  additional onPress, you would want to pass the value as well, but I can remove that if it is wrong.

For the `onValueChange && value` conditional, this is already assumed because we are initializing `onValueChange` so it IS there and we would have thrown if realValue was `null` or `undefined`.

Finally, isSelected is a simple comparison of non-null/non-undefined string values which we know exist because of `getValueForRadioButton`.  So we can compare them directly.  Consider also we want you to be able to compare empty strings as values in case for some reason your RadioButton is a reset.  So we do not need to verify a truthy string either.

peterpme (approved):
LGTM thanks for answering those questions!

PR #396, opened 2022-01-28
Fix checkboxRow mapping to 'checkboxRowValue'

smlarkin:
Per [P-2720](https://linear.app/draftbit/issue/P-2720/checkbox-and-checkbox-row-cause-redscreen), this will stop the Builder from breaking when both a CheckBox & CheckBoxRow are on the screen.


dhruska (approved):
👍

peterpme (approved):

PR #400, opened 2022-02-01
Bugfix: Allow defaultValue & value prop updates in NumberInput

smlarkin:
This is a fix of the oversight when the `NumberInput` was rewritten here...

https://github.com/draftbit/react-native-jigsaw/pull/398/files

...did not test in all cases.  This PR looks to be "the one to rule them all" for `NumberInput`.

### DIFF
--- a/packages/core/src/components/NumberInput.tsx
+++ b/packages/core/src/components/NumberInput.tsx
@@ -50,6 +50,8 @@ const NumberInput: FC<Props> = ({
     onChangeText?.(number);
   };
 
+  /* set currentStringNumberValue directly to defaultValue if it exists on load
+  ( no need to use TextInput's defaultValue because its value is always controlled & set ) */
   useEffect(() => {
     const defaultStringNumberValue = formatValueToStringNumber(
       defaultValue,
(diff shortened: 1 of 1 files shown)


packages/core/src/components/NumberInput.tsx
```diff
@@ -14,49 +14,75 @@ const NumberInput: FC<Props> = ({
...
+  useEffect(() => {
+    const defaultStringNumberValue = formatValueToStringNumber(
+      defaultValue,
+      currentStringNumberValue
+    );
+
+    if (currentStringNumberValue !== defaultStringNumberValue) {
+      setCurrentStringNumberValue(defaultStringNumberValue);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
adnelson, line 64: why don't we want the dependencies here? Seems like whenever they change we'd want to run this?

smlarkin: @adnelson you only set the defaultValue when you initialize the component.  It never goes back to this value.

Imagine if you set a default to 6 and the deleted the number, you would not want it set back to 6.

smlarkin: @adnelson BTW

`// eslint-disable-next-line react-hooks/exhaustive-deps` is used to suppress suppress if your eslint has the react hooks plugin.  I do.  If you checkout my branch and add all deps, you will see functionally these hooks will not work.

adnelson (approved):

PR #408, opened 2022-02-09
Fix Checkbox(Row, Group, GroupRow) and RadioRow issues

smlarkin:
Per @bluerssen `CheckboxRow` should function exactly the same as `Checkbox` with the added value of having a row of text horizontally adjacent to it.

This PR...

1. Fixes a flex issue with `RadioButtonRow` & `CheckboxRow`
2. Splits the original `CheckboxRow` 2 different implementations / components: 
  a. `CheckboxRow` which is now functionally the same as `Checkbox` with added props for a label, etc.
  b. `CheckboxGroupRow` which is the original implementation that is designed to work as a child of `CheckboxGroup` 
3. Refactors / all components -- usually just a little -- for clarity
4. Updated mappings for `CheckboxRow`
5. Edits the `CheckboxExample.js` file to reflect the new way to use `CheckboxRow`

Finally...

There is a lot of code duplication between `CheckboxRow` & `Checkbox` now.  They could be merged to a single component LATER :)

And the newly created `CheckboxGroupRow` component ... like `CheckboxGroup` is not being exported or used currently.

### DIFF
--- a/packages/core/src/components/Checkbox/CheckboxGroupRow.tsx
+++ b/packages/core/src/components/Checkbox/CheckboxGroupRow.tsx
@@ -84,13 +84,8 @@ const CheckboxGroupRow: React.FC<CheckboxGroupRowProps & IconSlot> = ({
 
   const handlePress = () => {
     if (!disabled) {
-      console.log("isChecked BEFORE", isChecked);
-      console.log("value", value);
-
       onPress?.(!isChecked);
       onValueChange?.(value, !isChecked);
-
-      console.log("isChecked AFTER", isChecked);
     }
   };
 
--- a/packages/core/src/components/Checkbox/CheckboxGroupRow.tsx
+++ b/packages/core/src/components/Checkbox/CheckboxGroupRow.tsx
@@ -1,0 +1,157 @@
+import * as React from "react";
+import {
+  StyleProp,
+  ViewStyle,
+  StyleSheet,
+  TextStyle,
+  View,
+  Platform,
+} from "react-native";
+import Checkbox, { CheckboxProps } from "./Checkbox";
+import Text from "../Text";
+import { useCheckboxGroupContext } from "./context";
+import type { IconSlot } from "../../interfaces/Icon";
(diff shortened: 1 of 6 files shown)


packages/core/src/mappings/CheckboxRow.ts
```diff
@@ -15,8 +18,13 @@ export const SEED_DATA = {
   layout: {
     minHeight: 50,
   },
-  triggers: [Triggers.OnPress],
+  triggers: [Triggers.OnPress, Triggers.OnCheck, Triggers.OnUncheck],
```
dhruska, line 21: 👍 nice

dhruska (approved):

packages/core/src/components/Checkbox/CheckboxGroupRow.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const handlePress = () => {
+    if (!disabled) {
+      console.log("isChecked BEFORE", isChecked);
+      console.log("value", value);
+
+      onPress?.(!isChecked);
+      onValueChange?.(value, !isChecked);
+
+      console.log("isChecked AFTER", isChecked);
```
dhruska, line 93:
```suggestion
      onPress?.(!isChecked);
      onValueChange?.(value, !isChecked);
```

packages/core/src/components/Checkbox/CheckboxGroupRow.tsx
```diff
@@ -0,0 +1,157 @@
...
+    onValueChange,
+    direction: parentDirection,
+  } = useCheckboxGroupContext();
+
+  const values = Array.isArray(selectedValues) ? selectedValues : [];
+  const isChecked = status || values.includes(value);
+
+  const handlePress = () => {
+    if (!disabled) {
+      console.log("isChecked BEFORE", isChecked);
```
peterpme, line 87: :bomb: these comments and i think its good

peterpme (approved):
LGTM!

PR #421, opened 2022-03-04
Fix Surface backgroundColor prop & unify overflow behavior

smlarkin:
Per [P-2780](https://linear.app/draftbit/issue/P-2780/surface-no-longer-respects-color-prop), users need to be able to use the backgroundColor prop.  And if overflow should act the same across all platforms. See image below...

Screen Shot 2022-03-04 at 8 43 12 AM


peterpme (approved):
💯

PR #423, opened 2022-03-07
Adjust ScreenContainer props...

smlarkin:
...these now match the new toggles added to `ScreenContainer` for `SafeArea`

### DIFF
--- a/packages/core/src/components/ScreenContainer.tsx
+++ b/packages/core/src/components/ScreenContainer.tsx
@@ -14,18 +14,18 @@ import type { Theme } from "../styles/DefaultTheme";
 type Props = {
   hasSafeArea: boolean;
   scrollable: boolean;
-  hasBottomSafeArea?: boolean;
-  hasTopSafeArea?: boolean;
+  hasBottomSafeArea: boolean;
+  hasTopSafeArea: boolean;
   theme: Theme;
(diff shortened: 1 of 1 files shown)


dhruska (dismissed):

packages/core/src/components/ScreenContainer.tsx
```diff
@@ -14,18 +14,18 @@ import type { Theme } from "../styles/DefaultTheme";
...
+  hasBottomSafeArea: boolean;
+  hasTopSafeArea: boolean;
   theme: Theme;
   style?: StyleProp<ViewStyle>;
   children?: React.ReactNode;
 };
 
 function ScreenContainer({
-  hasSafeArea = false,
+  hasSafeArea,
```
dhruska, line 25: If you're removing the default for `hasSafeArea`, is the type `boolean` correct? It won't always be supplied by the builder, right? I think perhaps it should be `hasSafeArea?: boolean;`

dhruska (approved):
LGTM 👍

packages/core/src/components/ScreenContainer.tsx
smlarkin: @dhruska I made it required with a default prop.  My feeling is if you use a default prop then you should not make it optimal.

PR #424, opened 2022-03-09
Fix Stepper so defaultValue does not break onChange callback

smlarkin:
As the title says ↑ this now works and the code is much simpler / refactored.

The ...

// eslint-disable-next-line react-hooks/exhaustive-deps

...is the right thing here.  More of those to come in the future.


peterpme (approved):
I'm assuming you tested on ios/android/web etc if so LGTM! 💯

smlarkin:
> I'm assuming you tested on ios/android/web etc if so LGTM! 💯

1 ) Yes

2) What else is there ... like etc = ? ... Just sayin'

PR #428, opened 2022-03-29
Optimize video chat in a WebView

smlarkin:
This is to make it easy to use a `WebView` with and external Video Chat Web App such as https://whereby.com/

If the `optimizeVideoChat` is toggled on the Builder, this service (and orthers) will play well in most modes.

There is a CORS error in PWA preview with some services.  There may be a followup PR for this behavior.


peterpme (approved):
LGTM with the assumption that you were able to test this on snack (web, android, ios) and then locally within the example repo

Nice work !!

PR #433, opened 2022-04-07
Add Banner component

dhruska:
Adds new Banner component

image

### DIFF
--- a/packages/core/src/components/Banner.tsx
+++ b/packages/core/src/components/Banner.tsx
@@ -22,7 +22,7 @@ type Props = {
   dismissable: boolean;
   buttonColor?: string;
   icon?: string;
-  content: string;
+  content?: string;
   contentStyle?: StyleProp<ViewStyle>;
   style?: StyleProp<ViewStyle>;
   ref?: React.RefObject<View>;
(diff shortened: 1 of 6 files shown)


packages/core/src/components/Banner.tsx
```diff
@@ -0,0 +1,204 @@
...
+import type { IconSlot } from "../interfaces/Icon";
+import shadow from "../styles/shadow";
+import { withTheme } from "../theming";
+import type { Theme } from "../styles/DefaultTheme";
+
+const ELEVATION = 1;
+const DEFAULT_MAX_WIDTH = 960;
+
+type Props = {
+  initiallyVisible: boolean;
```
smlarkin, line 21: For non-optional props, can you set a default when you destructure below?

packages/core/src/components/Banner.tsx
```diff
@@ -0,0 +1,204 @@
...
+      x: number;
+      y: number;
+      width: number;
+      height: number;
+    };
+  };
+};
+
+const Banner: React.FC<Props> = ({
+  initiallyVisible,
```
smlarkin, line 47: Not passing in a value might work, but can you be explicit about non-optional props and set to false?

smlarkin (changes requested):
I understand you are setting default props in the map, but wish that at the component level they were set too.  Up to you, of course.

smlarkin (dismissed):

packages/core/src/components/Banner.tsx
dhruska: Definitely

packages/core/src/components/Banner.tsx
dhruska: Certainly, that makes sense

smlarkin (approved):

PR #435, opened 2022-04-14
Add maps components

dhruska:
Adds MapView, MapCallout, MapMarker. Supports native and web.


packages/maps/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
peterpme, line 1: wondering if we need this

packages/maps/src/components/MapView.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from "react";
+import NativeMapView from "./ReactNativeMaps";
+import { MapViewProps } from "@draftbit/types";
+
+// Dynamically import from ./ReactNativeMaps so that we don't
```
peterpme, line 5: is this a TODO or is this what's happening?

packages/web-maps/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
peterpme, line 1: same

packages/maps/src/components/MapView.tsx
dhruska: Yeah, that's what's happening

packages/ui/package.json
```diff
@@ -43,8 +43,8 @@
   },
   "homepage": "https://github.com/draftbit/react-native-jigsaw#readme",
   "dependencies": {
-    "@draftbit/core": "^43.5.0",
-    "@draftbit/native": "^43.4.19"
+    "@draftbit/core": "^43.6.0-alpha.6",
```
peterpme, line 46: is this right? maps seems to be 43.5.0?

packages/web-maps/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@draftbit/web-maps",
+  "version": "43.6.0-alpha.6",
```
peterpme, line 3: are these versions right?

packages/maps/babel.config.js
dhruska: Looks like we don't, good call!

packages/ui/package.json
dhruska: That got pushed during my testing, just reverted

peterpme (approved):

PR #440, opened 2022-04-20
Fix default title value for Link component

dhruska:
* We were initializing the Link component with a `children` prop, when it really needs a `title` prop.
* Fixes unexpected flex alignment behavior

### DIFF
--- a/packages/core/src/components/Text.tsx
+++ b/packages/core/src/components/Text.tsx
@@ -45,7 +45,7 @@ export const BaseLink = ({
   return (
     <Text
       hitSlop={8}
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Text.tsx
```diff
@@ -45,7 +45,7 @@ export const BaseLink = ({
   return (
     <Text
       hitSlop={8}
-      style={[{ color: theme.colors.primary, alignSelf: "flex-start" }, style]}
```
dhruska, line 48: This causes unexpected behavior when the parent of this component has `alignItems: "center"` set (P-2663)

smlarkin (approved):

PR #446, opened 2022-04-29
Support animating to a location in MapView

dhruska:
Adds support for a class function to animate to a given location + zoom in MapView (native + web support).


adnelson (approved):
Sweet!

PR #449, opened 2022-05-03
Add new props to show user current location pin and set map region to user location

dhruska:
Adds two new props that allow users to show their users' current location on the map.

### DIFF
--- a/packages/maps/src/components/MapView.tsx
+++ b/packages/maps/src/components/MapView.tsx
@@ -1,5 +1,7 @@
 import * as React from "react";
-import NativeMapView from "./ReactNativeMaps";
+import * as Location from "expo-location";
+import Svg, { Ellipse } from "react-native-svg";
+import NativeMapView, { Marker } from "./ReactNativeMaps";
 import { MapViewProps } from "@draftbit/types";
 
 // Dynamically import from ./ReactNativeMaps so that we don't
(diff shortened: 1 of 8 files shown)


packages/maps/src/components/MapView.tsx
```diff
@@ -17,13 +19,42 @@ function zoomToAltitude(zoom: number) {
...
+      if (!this.props.showUserLocation) {
+        return;
+      }
+
+      let { status } = await Location.requestForegroundPermissionsAsync();
+      if (status !== "granted") {
+        return;
+      }
+
+      let location = await Location.getCurrentPositionAsync({});
```
adnelson, line 53: I believe these can both be `const` because they're not reassigned

```suggestion
      const { status } = await Location.requestForegroundPermissionsAsync();
      if (status !== "granted") {
        return;
      }

      const location = await Location.getCurrentPositionAsync({});
```

packages/web-maps/src/components/MapView.tsx
```diff
@@ -20,6 +30,22 @@ class MapView extends React.Component<MapViewProps, State> {
...
+        return;
+      }
+
+      let { status } = await Location.requestForegroundPermissionsAsync();
+      if (status !== "granted") {
+        return;
+      }
+
+      let location = await Location.getCurrentPositionAsync({});
+      this.setState({ userLocation: location.coords });
```
adnelson, line 45: `let` => `const`, also just a suggestion to use the same pattern for unpacking the record fields

```suggestion
      const { status } = await Location.requestForegroundPermissionsAsync();
      if (status !== "granted") {
        return;
      }

      const  { coords } = await Location.getCurrentPositionAsync({});
      this.setState({ userLocation: coords });
```

adnelson (dismissed):
A couple minor stylistic suggestions (take 'em or leave 'em), LGTM otherwise!

packages/maps/src/components/MapView.tsx
dhruska: Good call!

packages/web-maps/src/components/MapView.tsx
dhruska: Good call!

adnelson (approved):

PR #454, opened 2022-05-05
Support for data-driven Map Markers and Swiper Items

dhruska:
Adds a `data`, `renderItem`, and `keyExtractor` API for `MapView` and `Swiper` to allow for making their children data-driven, since `FlatList` is not compatible.

IMG_4483

IMG_4482

### DIFF
--- a/packages/web-maps/src/components/MapView.tsx
+++ b/packages/web-maps/src/components/MapView.tsx
@@ -19,8 +19,8 @@ type State = {
   };
 };
 
-class MapView extends React.Component<MapViewProps, State> {
-  constructor(props: MapViewProps) {
+class MapView extends React.Component<MapViewProps<any>, State> {
+  constructor(props: MapViewProps<any>) {
     super(props);
     this.state = {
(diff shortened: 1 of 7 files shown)


packages/web-maps/src/components/MapView.tsx
```diff
@@ -95,6 +95,9 @@ class MapView extends React.Component<MapViewProps, State> {
       scrollEnabled = true,
       mapType = "standard",
       style,
+      data,
```
adnelson, line 98: Should we give this a different name, like `points` or something? Just thinking that `data` isn't very descriptive

dhruska: Good call... how about `markersData`, to clarify that it's the data for the `MapMarker`s?

dhruska: I think for `Swiper`, `data` is probably ok unless there's another name that comes to mind for you

adnelson (approved):
🚀

PR #458, opened 2022-05-11
 Set full width as default for WebView

smlarkin:
Per [P-2843](https://linear.app/draftbit/issue/P-2843/webview-doesnt-show-on-mobile-previews), the WebView should always display full width when added to a screen.  It should also respond to height and width % and px settings.

NOTE ... unlike other width's we have to start with the window.width and base our settings from there with WebView.

### DIFF
--- a/packages/native/src/components/WebView.tsx
+++ b/packages/native/src/components/WebView.tsx
@@ -90,6 +90,16 @@ const NativeWebView: React.FC<WebViewProps> = ({
     );
   };
 
+  const getFinalWidth = () => {
+    if (typeof style?.width === "number") {
+      return style.width;
+    } else if (typeof style?.width === "string" && style.width.includes("%")) {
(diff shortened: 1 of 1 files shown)


packages/native/src/components/WebView.tsx
```diff
@@ -90,6 +90,16 @@ const NativeWebView: React.FC<WebViewProps> = ({
     );
   };
 
+  const getFinalWidth = () => {
+    if (typeof style?.width === "number") {
+      return style.width;
+    } else if (typeof style?.width === "string" && style.width.includes("%")) {
+      return width * (Number(style.width.replace("%", "")) / 100);
```
adnelson, line 97: Can we move the definition of `width` into this function? I think it would make it a little easier to understand.

smlarkin: @adnelson done.

PR #464, opened 2022-05-20
Rewrite Picker so controls work in Builder + add missing props

smlarkin:
Per [P-2846](https://linear.app/draftbit/issue/P-2846/changing-height-of-picker-component-messes-up-the-internal-picker) ... AND chats with Rahul, Amit, George, Brian ...

I've rewritten Picker to work as expected.  All props should continue to work AND there are 2 new ones... iconColor, iconSize...

ONLY ISSUE ... height set to 100%* (or any %)  is not predictable.  Instead you must set it to number and then continue styling.  Setting it to 100% doesn't change anything.


packages/core/src/components/Picker/Picker.tsx
```diff
@@ -60,39 +123,305 @@ const Picker: React.FC<Props> = ({
...
+        </NativePicker>
+      ) : null}
+    </View>
   );
 };
 
 export default withTheme(Picker);
+
+const styles = StyleSheet.create({
+  marginsContainer: { alignSelf: "stretch" },
```
peterpme, line 386: i'm surprised prettier didn't format this to live below?

packages/core/src/components/Picker/Picker.tsx
```diff
@@ -60,39 +123,305 @@ const Picker: React.FC<Props> = ({
...
+        onPress={togglePickerVisible}
+        style={styles.touchableContainer}
+      >
+        {/* outsetContainer */}
+        <View
+          pointerEvents="none"
+          style={[
+            styles.outsetContainer,
+            stylesWithoutBordersAndMargins,
+            // @ts-expect-error ... rejects border-related { key: undefined }
```
peterpme, line 294: what's `{key: undefined}`? am i misunderstanding that?

smlarkin: @peterpme Removed comment / Added type / N/A issue.

packages/core/src/components/Picker/Picker.tsx
```diff
@@ -1,31 +1,71 @@
...
-      label: option,
+      label: String(option),
       value: String(option),
     }));
   }
 
-  if (options[0].label && options[0].value) {
-    return options.map((option) => {
+  if (
+    typeof options[0] === "object" &&
```
peterpme, line 62: you can use `import { isObject } from 'lodash'` here which could be a tiny bit safer

packages/core/src/components/Picker/Picker.tsx
```diff
@@ -60,39 +122,311 @@ const Picker: React.FC<Props> = ({
...
 
   const pickerOptions = placeholder
     ? [{ value: placeholder, label: placeholder }, ...normalizedOptions]
     : normalizedOptions;
 
+  const { viewStyles, textStyles } = extractStyles(style);
+
+  const { colors } = theme;
+
+  const borders = [
```
peterpme, line 135: this seems like something that should be extracted out similar to `extractStyles` just in case we need it somewhere else - wdyt?

smlarkin: @peterpme Done.

packages/core/src/components/Picker/Picker.tsx
smlarkin: @peterpme Done.

packages/core/src/components/Picker/Picker.tsx
```diff
@@ -60,39 +127,300 @@ const Picker: React.FC<Props> = ({
...
+            borderTopWidth: 1,
+            borderRightWidth: 1,
+            borderLeftWidth: 1,
+          }
+        : {}),
+      borderBottomWidth: 1,
+      borderColor: unstyledColor,
+      borderStyle: "solid",
+    },
+    // ...pick(viewStyles, borders),
```
peterpme, line 160: Is this required?

packages/core/src/utilities.ts
```diff
@@ -34,6 +34,47 @@ export function extractStyles(style: StyleProp<any>) {
...
+  "borderTopWidth",
+  "borderRightWidth",
+  "borderBottomWidth",
+  "borderLeftWidth",
+  "borderColor",
+  "borderStyle",
+];
+
+export const marginStyleNames = [
+  "marginLeft",
```
peterpme, line 52: marginHorizontal/marginVertical?

packages/core/src/components/Picker/Picker.tsx
smlarkin: @peter Removed.

packages/core/src/utilities.ts
smlarkin: @peterpme added all margin and border styles.

peterpme (approved):
LGTM!

PR #482, opened 2022-07-12
Update AudioPlayer.tsx to handle audio finished

whw:
This change should reset the audio player after the source has finished playing.

In this context, reset means:
- Unload Audio
- Set slider position to zero
- Show the play icon

My hope is to inspire the spirit of the change - please feel free to edit/change as I didn't see any tests I could run this against.


peterpme:
Hi @whw sorry for the delay and thank you! We'll consider this internally and go from there :)

bluerssen:
Hey @whw - We updated Audio Component in a different PR to handle this, so closing. https://github.com/draftbit/react-native-jigsaw/pull/504/files Thank you!

whw:
@bluerssen awesome, thanks so much for letting me know! I doubt this is the right place but I'd like to put a small request in here: the ability to pass in styles for the audio player as well.

PR #488, opened 2022-08-09
SVG Component

bluerssen:
SVG Component!


bluerssen:
I think this is GTG but let me know what you think.

dsebek (approved):

PR #494, opened 2022-08-18
adds webview component

peterpme:


packages/webview/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@draftbit/web-view",
```
adnelson, line 2: Can we rename this folder to `packages/web-view` for consistency?

bluerssen:
@adnelson @peterpme we're closing this, I think, right?

PR #495, opened 2022-08-22
Fix disabling TextInput on iOS and Android

dsebek:
- Added `TEXT_FIELD_PROPS` for **TextField** copied from `TEXT_INPUT_PROPS` which it was using
- Updated **TextField** to use `TEXT_FIELD_PROPS` instead of `TEXT_INPUT_PROPS`
- Removed `disabled` prop from `TEXT_INPUT_PROPS` and `NUMBER_INPUT_PROPS`
- Updated default value of `editable` to `true` for `TEXT_INPUT_PROPS` and `NUMBER_INPUT_PROPS`
- Updated **TextInput** and **NumberInput** examples to use `editable` instead of `disabled`
- Added empty lines for visual separation of example components

Tested using example app on web and physical device


packages/types/src/component-types.ts
```diff
@@ -577,7 +566,7 @@ export const TEXT_INPUT_PROPS = {
   },
 };
 
-export const NUMBER_INPUT_PROPS = {
+export const TEXT_FIELD_PROPS = {
```
adnelson, line 569: If these are only now used in one place (the mappings file for `TextInput` and `TextField` then I'd say let's move them in there to avoid the complicated structure of having to go look them up in a different file.

Otherwise they're fine here.

dsebek: I moved them, but I think I'm maybe missing another reference somewhere

packages/core/src/mappings/TextInput.ts
```diff
@@ -4,12 +4,391 @@ import {
...
+      "If true, all the text will automatically be selected on focus",
+    editable: true,
+    required: false,
+    defaultValue: null,
+    formType: FORM_TYPES.boolean,
+    propType: PROP_TYPES.BOOLEAN,
+  },
+};
+
+const NUMBER_INPUT_PROPS = {
```
adnelson, line 225: Not to pile more work on you, but can we move this and the SEED_DATA for the NumberInput component to a new file `NumberInput.ts`?

dsebek: Sure, I was planning to do it in a subsequent PR actually, but now's fine

dsebek: Done

adnelson (approved):

PR #496, opened 2022-08-22
split webview into its own package

adnelson:
Make a separate package `@draftbit/webview` so we don't import webview automatically with `@draftbit/ui`


bluerssen:
@adnelson should we do this? i'm for it.

adnelson:
No longer needed

PR #527, opened 2022-10-17
HTML Element Mapping

gunjankothari:


bluerssen:
- Rename file from Elements to HtmlElements
- We need to add to the Index so they are exported, right?
- I dont think `BLOCK_STYLES_SECTIONS` is sufficient here, because most of these are text elements that need typography, so for those text elements we should provide the same as text components. 
-

bluerssen (changes requested):
See my other comment

bluerssen:
Ready to be merged when @adnelson fixes bug re: mappings without props

bluerssen (approved):

YoussefHenna:
Closed in favor of https://github.com/draftbit/react-native-jigsaw/pull/591

PR #534, opened 2022-11-30
introduced onIndexChanged event

gunjankothari:


bluerssen:
A Draftbit-side change has to be implemented before we can merge.

PR #544, opened 2022-12-15
Fix Date Picker Styling

bluerssen:


bluerssen:
@gunjankothari This is stuff moved over from your other PR. Only thing we need to do is to add in web related similar styles to the ones we added in this for iOS/android

PR #550, opened 2022-12-22
Datepicker viewStyle fix

gunjankothari:


YoussefHenna:
I don't see why this is needed. Why should `viewStyle` be applied to the TextInput. Can we close this?

PR #556, opened 2023-01-06
Add LongPress and other props to buttons and pressables

bluerssen:
- Adding LongPress and other props to Buttons and Touchables/Pressables

### DIFF
--- a/packages/core/src/components/Pressable.tsx
+++ b/packages/core/src/components/Pressable.tsx
@@ -1,5 +1,9 @@
 import React from "react";
-import { Pressable, PressableProps, ViewStyle } from "react-native";
+import {
+  Pressable as NativePressable,
+  PressableProps,
+  ViewStyle,
+} from "react-native";
 
 type Props = {
   disabled?: boolean;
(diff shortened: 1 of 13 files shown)


packages/core/src/components/Pressable.tsx
```diff
@@ -0,0 +1,50 @@
...
+type Props = {
+  disabled?: boolean;
+  children: React.ReactNode;
+  style?: ViewStyle;
+  onPress?: () => void;
+  onLongPress?: () => void;
+  delayLongPress?: number;
+  hitSlop?: number;
+  activeOpacity?: number;
+  disabledOpacity?: number;
```
adnelson, line 17: Can we omit the ones of these which are already covered by `NativePressable`? I think for example `onPress`, `onLongPress`, `children` etc. This would allow us to see at a glance what this wrapper was adding, which is good info especially if we're going to phase out jigsaw eventually.

bluerssen: I tried to slim it down and got to Style, activeOpacity, and disabledopacity in the types.

PR #598, opened 2023-02-15
Add Table Component

YoussefHenna:
I have refactored this component heavily from https://github.com/draftbit/react-native-jigsaw/pull/346 to what I feel is best suited for draftbit.
- 3 components: Table -> TableRow -> TableCell
- The table components are now merely components that facilitate laying out the views and border drawing. Makes it highly customizable and can hold anything, not only Text
- Props allow customization. This customization is passed from parent to child (Table -> TableRow for example) by default, but each child can then override them.
- Added renderItem functionality for rows
- Removed TableHeader component. TableRow can be marked as a header, which just changes the background to the theme color and sticks it to the top when scrolling.
  - A header row can be inside the renderItem as well. Only the first header is rendered as a header and the rest are ignored. This allows usage in Draftbit where all children are automatically placed in the renderItem

Screenshot_1676452884


packages/core/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,176 @@
...
+  style?: StyleProp<ViewStyle>;
+  theme: Theme;
+}
+
+const Table = <T extends object>({
+  theme,
+  borderWidth = 1,
+  borderColor = theme.colors.divider,
+  borderStyle = "solid",
+  drawTopBorder = true,
```
bluerssen, line 32: I guess we cant just use border styling props here?

YoussefHenna: I initially had it like this because these props are passed down to children (Row and Cell) and used there for styling as well. But definitely makes sense to use normal styling, and then pass the values from there. I'll see if it'll cause any problems

bluerssen (approved):

PR #599, opened 2023-02-15
Added support for renderItem in DeckSwiper

YoussefHenna:
- Made DeckSwiper work when provided `data` and `renderItem` instead of children.
- Direct children still works if these not provided


YoussefHenna:
Tested with data on Draftbit and works as expected.

bluerssen (approved):

PR #342, opened 2021-10-27
P-2589 Bottom Sheet with example

gunjankothari:

PR #600, opened 2023-02-16
Add BottomSheet component

YoussefHenna:
- Modified changes from https://github.com/draftbit/react-native-jigsaw/pull/342
- Using `react-native-scroll-bottom-sheet` for both web and mobile instead of 2 libraries since this one already supports all platforms
- Copied source code of library (single file) into jigsaw in order to fix bugs it had
- Added more props + refactored the code
- Added `createArrayProp` function


packages/core/package.json
```diff
@@ -54,9 +54,11 @@
     "lodash.omit": "^4.5.0",
     "lodash.tonumber": "^4.0.3",
     "react-native-deck-swiper": "^2.0.12",
+    "react-native-gesture-handler": "^2.9.0",
```
bluerssen, line 57: The recommended version for these packages for expo 46 are:
    "react-native-gesture-handler": "~2.5.0",
    "react-native-reanimated": "~2.9.1"

The recommended versions for expo 47 are:
    "react-native-gesture-handler": "~2.8.0",
    "react-native-reanimated": "~2.12.0"

Is it possible to use these sets with this component? Otherwise I worry we'll run into conflicts when these apps need to be built in expo 46/47

packages/core/src/mappings/BottomSheet.ts
```diff
@@ -0,0 +1,78 @@
...
+} from "@draftbit/types";
+
+export const SEED_DATA = {
+  name: "Bottom Sheet",
+  tag: "BottomSheet",
+  description:
+    "A draggable Bottom sheet that snaps to specific points. Renders children in ScrollView",
+  category: COMPONENT_TYPES.bottomsheet,
+  stylesPanelSections: CONTAINER_COMPONENT_STYLES_SECTIONS,
+  layout: { paddingHorizontal: 16, paddingVertical: 10 },
```
bluerssen, line 19: Does it make sense to have these defaults here or just in the component itself?

YoussefHenna: I felt like having them here clearly shows the user the default applied padding. But then again, "auto" is a good representation of the default padding. I'll remove it from the mapping.

packages/core/package.json
YoussefHenna: I remember trying these versions initially and running into issues on draftbit. However, just tried them now, and everything works. In fixing other issues, these were probably fixed as well.

bluerssen (approved):

PR #619, opened 2023-03-06
Add Triggers and Custom markers to Maps

YoussefHenna:
- Added onRegionChange trigger that us triggered when panning to a new region
- Added custom marker ability through `pinImageUrl` prop
- Added `onPress` trigger to marker


bluerssen (dismissed):

YoussefHenna:
I added the ability to add an image as the marker. This is pending Rahul's response on Linear to see if this is needed/useful or not.

bluerssen (approved):

PR #636, opened 2023-03-20
Fix `FlatList` not working because of the forced `flex` style attribute

needs:
Some testing in Snack showed that the `FlatList` component was not properly working, the main reason being that the `flex` style attribute was forced to `1` in the `FlatList` component.  For whatever reason, when this settings is set, either the Webpreview fail to render the FLatlist, or the Appetize preview do.


YoussefHenna (approved):
Looks good!

PR #668, opened 2023-04-30
Update disableXSwipe prop to be consistent with swipe directions

YoussefHenna:
- `disableLeftSwipe` refers to swipe from right -> left and not the other way around
- same for `disableRightSwipe`


packages/core/src/components/SwipeableItem/SwipeableItem.tsx
```diff
@@ -242,6 +244,8 @@ const SwipeableItem: React.FC<React.PropsWithChildren<Props>> = ({
         swipeGestureEnded={onStopSwiping}
         closeOnRowPress={closeOnPress}
         friction={friction}
+        disableLeftSwipe={disableRightSwipe}
```
bluerssen, line 247: this seems kinda confusing...

bluerssen (approved):

PR #717, opened 2023-06-14
Multiple minor changes

YoussefHenna:
- Add `pause` and `play` methods/actions to AudioPlayer and VideoPlayer
- Extract children of components when nested in `<></>` to ensure that the `.type` still works even when nested in a react fragment. This can happen in Draftbit when wrapped in a conditional for example.
- Added `initialTabIndex` prop to tab view

Closes https://linear.app/draftbit/issue/P-3697/audio-and-video-players
Closes https://linear.app/draftbit/issue/P-3694/not-able-to-hide-unhide-tabview-tab
Closes https://linear.app/draftbit/issue/P-3692/action-to-change-active-index-of-tab-view

### DIFF
--- a/packages/core/src/components/SwipeableItem/SwipeableItem.tsx
+++ b/packages/core/src/components/SwipeableItem/SwipeableItem.tsx
@@ -12,6 +12,7 @@ import {
   extractBorderAndMarginStyles,
   extractEffectStyles,
   extractFlexItemStyles,
+  extractIfNestedInFragment,
   extractPositionStyles,
   extractSizeStyles,
   extractStyles,
@@ -74,7 +75,7 @@ type Props = SwipeableItemProps & RightSwipeProps & LeftSwipeProps;
 const SwipeableItem: React.FC<React.PropsWithChildren<Props>> = ({
   theme,
   style,
-  children,
+  children: childrenProp,
   Icon,
   closeOnPress,
   leftOpenValue,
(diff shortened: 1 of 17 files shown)


packages/ui/package.json
```diff
@@ -43,9 +43,9 @@
   },
   "homepage": "https://github.com/draftbit/react-native-jigsaw#readme",
   "dependencies": {
-    "@draftbit/core": "48.2.2",
-    "@draftbit/native": "48.2.2",
-    "react-native-reanimated": "~2.12.0"
+    "@draftbit/core": "48.2.3",
+    "@draftbit/native": "48.2.3",
+    "react-native-reanimated": "~2.14.4"
```
YoussefHenna, line 48: This was a version mismatch caused by an older commit. This is the proper version for expo 48.

packages/core/src/components/SwipeableItem/SwipeableItem.tsx
```diff
@@ -124,9 +125,18 @@ const SwipeableItem: React.FC<React.PropsWithChildren<Props>> = ({
   const [componentWidth, setComponentWidth] = React.useState<number | null>(
     null
   );
+
+  const children: React.ReactNode[] = React.useMemo(
```
adnelson, line 129: Can we get a comment explaining why we need to pull extract components out of fragments in the children prop?

adnelson (commented):
Looks great, you could add a comment in one spot

PR #799, opened 2023-10-04
fix: ignore textTransform uppercase label of TabView

sieuhuflit:
## Problem

`react-native-tab-view/TabBarItem.tsx` force label into uppercase 

Screenshot 2023-10-04 at 17 30 42

## Overview

- Add textTransform to none to prevent uppercase of Tab view title

## Demo

Screenshot 2023-10-04 at 17 23 32

Closes https://linear.app/draftbit/issue/P-4149/not-able-to-make-capitalize-tab-view-title

### DIFF
--- a/packages/core/src/components/TabView/TabView.tsx
+++ b/packages/core/src/components/TabView/TabView.tsx
@@ -111,7 +111,7 @@ const TabViewComponent: React.FC<React.PropsWithChildren<TabViewProps>> = ({
         indicatorStyle={{
           backgroundColor: indicatorColor || theme.colors.primary,
         }}
-        labelStyle={textStyles}
+        labelStyle={[textStyles, { textTransform: "none" }]}
         renderIcon={({ route, color }) =>
           route?.icon ? (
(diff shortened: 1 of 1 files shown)


packages/core/src/components/TabView/TabView.tsx
```diff
@@ -111,7 +111,7 @@ const TabViewComponent: React.FC<React.PropsWithChildren<TabViewProps>> = ({
         indicatorStyle={{
           backgroundColor: indicatorColor || theme.colors.primary,
         }}
-        labelStyle={textStyles}
+        labelStyle={[textStyles, { textTransform: "none" }]}
```
YoussefHenna, line 114:
```suggestion
        labelStyle={[{ textTransform: "none" }, textStyles]}
```
I'd reverse the order, so the user is still able to override it if needed

YoussefHenna (changes requested):
Thanks for fixing this, just a minor comment

YoussefHenna (approved):

PR #800, opened 2023-10-12
Add prop to improve performance of MapView

YoussefHenna:
- When the `tracksViewChanges` prop on markers is enabled, every marker on the screen keeps track of changes in the parent view. When there are many markers, this can create a sluggish and laggy map when trying to drag/pan (most noticeable on android).
  - This prop is enabled by default - https://github.com/react-native-maps/react-native-maps/blob/master/docs/marker.md#props 
- This PR temporarily disables the prop when dragging then enables it after dragging stops
- Disabling entirely creates another issue where markers and clusters are never updated, so this is best solution to prevent laggy panning while still allowing markers and clusters to be able to update.

### DIFF
--- a/packages/maps/src/__tests__/MapView.test.tsx
+++ b/packages/maps/src/__tests__/MapView.test.tsx
@@ -186,6 +186,7 @@ describe("MapView tests", () => {
               }
             }
             onPress={[Function]}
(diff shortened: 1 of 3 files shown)


adnelson (approved):
Makes sense! Good stuff :)

PR #881, opened 2024-05-15
feat: support mode modal for picker component

sieu-db:
## Overview

Support new mode is modal picker

- Prevent the issue concerning zIndex conflicts.
- Introduces an enhancement to dismiss the picker by tapping outside the modal. This is crucial as it aligns with the standard user experience of modal operation.
- Adds a feature to customize the backdrop overlay of the modal. This help user can fully customize the overlay, to make the UI behavior exactly the same as dropdown picker mode
- ❌ This mode not supported multipicker

## Existing problem

We have some problem if picker got wrap by the View

Screenshot 2024-05-15 at 21 18 30

## Demo

https://github.com/draftbit/react-native-jigsaw/assets/159281348/0f660a9a-4736-434d-90ef-feb2fdc40e53

### DIFF
--- a/example/src/PickerExample.tsx
+++ b/example/src/PickerExample.tsx
@@ -57,6 +57,64 @@ function PickerExample() {
         />
       </Picker>
 
+      <Section style={{}} title="Picker - Modal">
+        <></>
+      </Section>
+      <Picker
+        label="Make"
+        placeholder="Select a make..."
+        options={OPTIONS}
+        value={value1}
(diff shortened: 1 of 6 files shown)


YoussefHenna (approved):
Awesome stuff! One thing is that there is bit of duplicated code from the dropdown picker. I'll merge this to a separate branch and de-deplicate before I merge to master. But otherwise, great work!

PR #890, opened 2024-06-10
Allow for `null` behavior with `KeyboardAvoidingView`

needs:
Sieu needs it for an app.  Types says that `behavior` can either be some value or `undefined`, but Sieu uses `null` and it works, while using `undefined` doesn't.  Se the associated Linear task.

### DIFF
--- a/packages/core/src/components/KeyboardAvoidingView.tsx
+++ b/packages/core/src/components/KeyboardAvoidingView.tsx
@@ -8,7 +8,7 @@ import {
 const isIos = Platform.OS === "ios";
 const isAndroid = Platform.OS === "android";
 
-type KeyboardAvoidingViewBehavior = "height" | "position" | "padding";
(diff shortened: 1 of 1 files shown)


packages/core/src/components/KeyboardAvoidingView.tsx
```diff
@@ -51,7 +51,7 @@ const KeyboardAvoidingView: React.FC<KeyboardAvoidingViewProps> = ({
 
   return (
     <KeyboardAvoidingViewComponent
-      behavior={behaviorResult}
+      behavior={behaviorResult ?? undefined}
```
YoussefHenna, line 54: This is an optional value, so I think they can just not pass anything and it would be equivalent to passing null? Or am I missing something?

YoussefHenna (approved):

PR #909, opened 2024-07-29
add loading component

sieu-db:
## Overview
- Support AcitivtyIndicator component
- Make sure it's worked with existing size is "small" or "large", make sure it's scale correctly on iOS device

https://github.com/user-attachments/assets/98cee27f-5176-437a-95ed-764dae4619ad


packages/core/src/components/ActivityIndicator.tsx
```diff
@@ -0,0 +1,125 @@
...
+  Grid,
+  Plane,
+  Pulse,
+  Swing,
+  Wander,
+  Wave,
+} from "react-native-animated-spinkit";
+
+export enum ActivityIndicatorType {
+  plane = "plane",
```
YoussefHenna, line 27:
```suggestion
export enum ActivityIndicatorType {
  default = "default"
  plane = "plane",
```
Let's have a default type to make it more explicit

packages/core/src/components/ActivityIndicator.tsx
```diff
@@ -0,0 +1,125 @@
...
+const getScaleLevel = (size?: number | "small" | "large"): number => {
+  if (typeof size === "number") return size / smallActivityIndicatorSize;
+  return size === "large" ? 2 : 1;
+};
+
+const ActivityIndicator: React.FC<React.PropsWithChildren<Props>> = ({
+  color,
+  theme,
+  type,
+  size,
```
YoussefHenna, line 83:
```suggestion
  theme,
  color = theme.colors.branding.primary,
  type = "default",
  size = "small",
```
Let's add default values here to be consistent with how it's in other components

packages/core/src/components/ActivityIndicator.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  // This is due to limitations with iOS UIActivityIndicator
+  // Refer: https://github.com/facebook/react-native/issues/12250
+  const scaleLevel = getScaleLevel(size);
+  if (Platform.OS === "ios") {
+    return (
+      <View
+        style={{
+          transform: [{ scale: scaleLevel }],
+        }}
```
YoussefHenna, line 102: The `style` prop should be applied here instead of the child component so we'd need to extract it from `rest` and apply it directly.

YoussefHenna (dismissed):
Some minor comments, otherwise awesome stuff!

YoussefHenna (approved):
Awesome. Don't forget to make the same changes for the `49` branch

sieu-db:
Yep sure. Thanks

PR #915, opened 2024-08-07
Add ref support to withTheme HOC

sieu-db:
## Overview
- Support forward ref for withtheme
- Update example of BottomSheet try to use some BottomSheet method

## Demo

https://github.com/user-attachments/assets/dc33f706-6575-420f-b5de-0365fa731fac

### DIFF
--- a/packages/theme/src/Provider.tsx
+++ b/packages/theme/src/Provider.tsx
@@ -115,13 +115,24 @@ const useChangeTheme = (): ((themeName: string) => void) => {
   return changeTheme;
 };
 
+// const withTheme = <Props extends { theme: ReadTheme }>(
+//   Component: React.ComponentType<Props>
+// ): React.ComponentType<Omit<Props, "theme">> => {
+//   return (props: Omit<Props, "theme">) => {
+//     const { theme } = React.useContext(ThemeContext);
+//     return <Component {...(props as Props)} theme={theme} />;
+//   };
+// };
+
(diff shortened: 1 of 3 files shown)


packages/theme/src/Provider.tsx
```diff
@@ -117,11 +117,13 @@ const useChangeTheme = (): ((themeName: string) => void) => {
...
 const withTheme = <Props extends { theme: ReadTheme }>(
   Component: React.ComponentType<Props>
-): React.ComponentType<Omit<Props, "theme">> => {
-  return (props: Omit<Props, "theme">) => {
-    const { theme } = React.useContext(ThemeContext);
-    return <Component {...(props as Props)} theme={theme} />;
-  };
+) => {
+  return React.forwardRef(
+    (props: Omit<Props, "theme">, ref: React.Ref<any>) => {
```
YoussefHenna, line 122: With components that need a forwarded ref, I usually don't wrap them in a `withTheme`, instead I get the theme by using the `useTheme` hook and just export them as is. See the `AudioPlayerWithInterface.tsx` for example. By doing that, the type of the ref can be maintained and is not `any` like it is here. This works as well though, it'll just lose the type

YoussefHenna (dismissed):

packages/theme/src/Provider.tsx
sieu-db: Thanks, adjust BottomSheet component

YoussefHenna (approved):

PR #920, opened 2024-08-08
FlashList padding and styles fix

YoussefHenna:
- Limit what is passed to flash list's styles based on https://shopify.github.io/flash-list/docs/usage/#contentcontainerstyle. Otherwise it shows a warning
- Convert any percent padding values of the flash list to be numbers, otherwise it leads to freezing and eventually crashing apps.

Closes https://linear.app/draftbit/issue/P-5282/page-freezes-when-padding-is-defined-for-flashlist-component
Closes https://linear.app/draftbit/issue/P-5294/warning-about-padding-for-flashlist

### DIFF
--- a/packages/core/src/components/SimpleStyleScrollables/useSplitContentContainerStyles.ts
+++ b/packages/core/src/components/SimpleStyleScrollables/useSplitContentContainerStyles.ts
@@ -57,3 +57,25 @@ export default function useSplitContentContainerStyles(
     };
   }, [originalStyle]);
 }
+
+export function useFlashListSplitContentContainerStyles(
+  originalStyle: StyleProp<ViewStyle>
+) {
+  const { style, contentContainerStyle } =
+    useSplitContentContainerStyles(originalStyle);
+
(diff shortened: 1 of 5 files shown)


packages/core/src/components/SimpleStyleScrollables/useSplitContentContainerStyles.ts
```diff
@@ -1,6 +1,9 @@
-import { StyleProp, ViewStyle, StyleSheet } from "react-native";
+import { StyleProp, ViewStyle, StyleSheet, Dimensions } from "react-native";
 import { pick, omit } from "lodash";
-import { useDeepCompareMemo } from "../../utilities";
+import { extractPercentNumber, useDeepCompareMemo } from "../../utilities";
+
+const DEVICE_WIDTH = Dimensions.get("window").width;
+const DEVICE_HEIGHT = Dimensions.get("window").width;
```
sieu-db, line 6: This should be height

PR #924, opened 2024-08-08
Support loading indicator component

sieu-db:


packages/core/src/components/LoadingIndicator.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const LoadingIndicator: React.FC<React.PropsWithChildren<Props>> = ({
+  theme,
+  color = theme.colors.branding.primary,
+  type = LoadingIndicatorType.plane,
+  size,
+  style,
+  ...rest
+}) => {
+  const spinnerColor = color ?? theme.colors.branding.primary;
```
YoussefHenna, line 66: This default is already done at the top with `color theme.colors.branding.primary`

YoussefHenna (dismissed):
👍 Need the same for the `49` branch

packages/core/src/components/LoadingIndicator.tsx
sieu-db: Thanks 🪨

YoussefHenna (approved):

PR #925, opened 2024-08-09
Support isLooping for audio player

sieu-db:
## Overview
- Support isLooping for audio player

## Demo

https://github.com/user-attachments/assets/dd824fa2-41d6-4a1f-9a02-db4d90e0e59b


packages/core/src/components/MediaPlayer/AudioPlayer/HeadlessAudioPlayer.tsx
```diff
@@ -97,6 +101,7 @@ const HeadlessAudioPlayer = React.forwardRef<
 
       const { sound } = await Audio.Sound.createAsync(finalSource);
       setCurrentSound(sound);
+      sound.setIsLoopingAsync(isLooping);
```
YoussefHenna, line 104: This will only set the value when the audio is reloaded (audio source changes). Maybe we should have a separate useEffect that updates the looping state

```tsx
React.useEffect(() => {
    currentSound?.setIsLoopingAsync(isLooping);
}, [isLooping])
```

packages/core/src/components/MediaPlayer/AudioPlayer/AudioPlayerWithInterface.tsx
```diff
@@ -31,6 +31,7 @@ const AudioPlayerWithInterface = React.forwardRef<
       hidePlaybackIcon = false,
       hideDuration = false,
       hideSlider = false,
+      isLooping = false,
```
YoussefHenna, line 34: This can just be passed along with `...rest`, since it's not used here.

YoussefHenna (changes requested):

packages/core/src/components/MediaPlayer/AudioPlayer/AudioPlayerWithInterface.tsx
sieu-db: Hmmm, it's pass over to HeadlessAudioPlayer of line 123, not sure about this one

YoussefHenna: `...rest` is also passed into the HeadlessAudioPlayer, so it'll get passed with those as well.

packages/core/src/components/MediaPlayer/AudioPlayer/AudioPlayerCommon.ts
```diff
@@ -20,4 +21,5 @@ export interface AudioPlayerInterfaceProps {
   hidePlaybackIcon?: boolean;
   hideDuration?: boolean;
   hideSlider?: boolean;
+  isLooping?: boolean;
```
YoussefHenna, line 24: Already defined in the HeadlessAudioPlayerProps

sieu-db:
Adjust this one 🙏 @YoussefHenna

packages/core/src/components/MediaPlayer/AudioPlayer/HeadlessAudioPlayer.tsx
YoussefHenna: We can remove this one now since we have the other useEffect @sieu-db 

```suggestion
```

YoussefHenna: Otherwise all good!

sieu-db: Yep, got it, thanks

PR #926, opened 2024-08-11
Add support for `onRefresh` on scroll views

YoussefHenna:
- This prop is available to list components but not scroll view, this adds support for it.


sieu-db (commented):
LGTM 💯

PR #934, opened 2024-08-18
Support timer component

sieu-db:
## Overview
- Add support for multiple time formats: "ss" | "mm:ss" | "hh:mm:ss" | "ss:ms" | "mm:ss:ms" | "hh:mm:ss:ms";
- Implement count-up and count-down timer functionality
- Can set interval & initial time also
- Introduce onTimerChange and onTimerEnd triggers
- Add refs for timer control: start(), stop(), reset(newTime?)

https://github.com/user-attachments/assets/9af1b5fd-5f3a-4ea7-a6a2-f82204c38e56


packages/core/src/components/Timer.tsx
```diff
@@ -0,0 +1,139 @@
...
+    },
+    ref
+  ) => {
+    const defaultInitialTime = countDirection === "up" ? 0 : 100000;
+    const [time, setTime] = useState(initialTime ?? defaultInitialTime);
+    const timerRef = useRef<NodeJS.Timeout | null>(null);
+
+    useEffect(() => {
+      onTimerChange?.(time);
+    }, [time, onTimerChange]);
```
YoussefHenna, line 45:
```suggestion
    }, [time]);
```
I would remove the callback from here. I know it will warn you otherwise, but in most cases this callback will be set as an anonymous function. Something like `onTimerChange={()=>{...}}`, which means it will cause the useEffect to be called on every screen re-render since a new anonymous function is created everytime. We had this issue with several components so far and had to change it. It won't cause any issues because the use effect will still use the updated onTimerChange everytime the `time` changes.

Add a `// eslint-disable-next-line react-hooks/exhaustive-deps` to ignore the warning.

packages/core/src/components/Timer.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+    const startTimer = () => {
+      if (timerRef.current) return;
+      timerRef.current = setInterval(() => {
+        setTime((prevTime) => {
+          const newTime =
+            countDirection === "up"
+              ? prevTime + updateInterval
+              : prevTime - updateInterval;
+          if (newTime <= 0 && countDirection === "down") {
```
YoussefHenna, line 61: So if the direction is `"up"`, the timer never ends? Maybe we can have a prop `timerEndTime` or something similar that tells us when a timer should be considered completed

packages/core/src/components/Timer.tsx
```diff
@@ -0,0 +1,139 @@
...
+      format = "mm:ss",
+      onTimerChange,
+      onTimerEnd,
+      countDirection = "up",
+    },
+    ref
+  ) => {
+    const defaultInitialTime = countDirection === "up" ? 0 : 100000;
+    const [time, setTime] = useState(initialTime ?? defaultInitialTime);
+    const timerRef = useRef<NodeJS.Timeout | null>(null);
```
YoussefHenna, line 41: This timer is never cleared on component unmount, which can cause a leak where we have timers running that do not belong to any component. I think we should use a `useEffect` somewhere with the cleanup callback. Something like this https://stackoverflow.com/a/63334017

YoussefHenna (changes requested):

YoussefHenna:
Another thought I had, maybe we can have a headless mode for this component, similar to how we have it for AudioPlayer. In case someone needs a timer without actually rendering anything. Not necessary for the first version though, we can add it later.

packages/core/src/components/Timer.tsx
sieu-db: Thanks, good point

packages/core/src/components/Timer.tsx
```diff
@@ -0,0 +1,139 @@
...
+          ).padStart(2, "0")}:${String(ms).padStart(3, "0")}`;
+        case "ss:ms":
+          return `${String(totalSeconds).padStart(2, "0")}:${String(
+            ms
+          ).padStart(3, "0")}`;
+        default:
+          return `${String(minutes).padStart(2, "0")}:${String(
+            seconds
+          ).padStart(2, "0")}`;
+      }
```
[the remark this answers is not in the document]
sieu-db: Yep sure, it cleaner

packages/core/src/components/Timer.tsx
sieu-db: Thanks, good point 👍

YoussefHenna (dismissed):
Awesome!

PR #935, opened 2024-08-18
Add selectable text component

sieu-db:
## Overview
- Add selectable text component

## Demo

https://github.com/user-attachments/assets/f6674e41-86b4-44a9-9e0f-dedb62f69a7f

### DIFF
--- a/packages/core/src/components/SelectableText.tsx
+++ b/packages/core/src/components/SelectableText.tsx
@@ -1,0 +1,50 @@
+import React from "react";
+import {
+  Platform,
+  TextInput,
+  Text,
+  TextStyle,
+  I18nManager,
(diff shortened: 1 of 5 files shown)


packages/core/src/components/SelectableText.tsx
```diff
@@ -0,0 +1,50 @@
...
+  I18nManager,
+  StyleSheet,
+  TextInputProps,
+  TextProps,
+  StyleProp,
+} from "react-native";
+
+type Props = {
+  style?: StyleProp<TextStyle>;
+  text?: string;
```
YoussefHenna, line 16: Can the text be passed in as `children`? Similar to how the normal `Text` component works

PR #941, opened 2024-08-25
Add a way to persist theme selection

YoussefHenna:
- When we eventually add themes to draftbit and an action to change the theme, we should also have a way to persist that selection so that the theme does not reset every time an app is relaunched. 

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit fa62274096f6e9b9d97808cbd02358e4fab377d6  | 
|--------|--------|

### Summary:
This PR adds theme persistence using AsyncStorage and updates jest setup files to mock AsyncStorage for testing.

**Key points**:
- Adds theme persistence using AsyncStorage in the Provider component.
- Updates jest setup files to mock AsyncStorage for testing.
- Adds `@react-native-async-storage/async-storage` to dependencies.
- Persists theme selection using `AsyncStorage` in `Provider.tsx`.
- Introduces `SAVED_SELECTED_THEME_KEY` constant for storage key.
- Modifies `changeTheme` function to accept `ChangeThemeOptions`.
- Adds `React.useEffect` to load persisted theme on initialization.
- Updates `useChangeTheme` hook to support `ChangeThemeOptions`.

----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)

### DIFF
--- a/packages/theme/src/Provider.tsx
+++ b/packages/theme/src/Provider.tsx
@@ -1,8 +1,16 @@
 import React from "react";
 import { Dimensions, Platform } from "react-native";
+import AsyncStorage from "@react-native-async-storage/async-storage";
 import createThemeValuesProxy from "./createThemeValuesProxy";
 import DefaultTheme from "./DefaultTheme";
-import { Breakpoints, ReadTheme, ValidatedTheme } from "./types";
+import {
+  Breakpoints,
+  ChangeThemeOptions,
+  ReadTheme,
+  ValidatedTheme,
+} from "./types";
+
+const SAVED_SELECTED_THEME_KEY = "saved_selected_theme";
 
 type ThemeContextType = {
   theme: ReadTheme;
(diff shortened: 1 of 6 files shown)


packages/theme/src/Provider.tsx
```diff
@@ -98,6 +103,30 @@ const Provider: React.FC<React.PropsWithChildren<ProviderProps>> = ({
...
+    const listener = Dimensions.addEventListener("change", ({ window }) =>
+      setDeviceWidth(window.width)
+    );
+    return () => {
+      listener.remove();
+    };
+  });
+
+  React.useEffect(() => {
+    const run = async () => {
```
draftbitbuildservices, line 116: Consider if the UI shakes slightly during initialization due to the absence of a wait state when loading the theme.

👉  Example: the UI may briefly display with theme1 before switching to theme2, causing a flicker.

If thorough testing shows no issue, you can disregard this comment.

sieu-db, line 116: Consider if the UI shakes slightly during initialization due to the absence of a wait state when loading the theme.

👉  Example: the UI may briefly display with theme1 before switching to theme2, causing a flicker.

If thorough testing shows no issue, you can disregard this comment.

YoussefHenna: I thought about the same thing, but saw no flicker during testing. We can always come back to it and fix if the issue does arise at some point.

PR #946, opened 2024-08-28
Adjust onSettle index of BottomSheet

sieu-db:
## Overview
- Figure out the problem that causing auto add 1 index in, as default it have enableDynamicSizing prop so it causing add index in, it causing mess up existing index logic we have
- Refer : https://gorhom.dev/react-native-bottom-sheet/props#enabledynamicsizing

Screenshot 2024-11-04 at 4 41 06 PM

https://github.com/user-attachments/assets/d594faca-b357-4c75-bcd7-89d56a2b2c12

----

> [!IMPORTANT]
> Fixes index calculation in `BottomSheet` by adjusting for `enableDynamicSizing` prop and updates example to prevent auto index increment.
> 
>   - **Behavior**:
>     - Fixes index calculation in `BottomSheet` by adjusting for `enableDynamicSizing` prop.
>     - Sets `enableDynamicSizing` to `false` in `BottomSheetExample.tsx` to prevent auto index increment.
>   - **Props**:
>     - Adds `enableDynamicSizing` prop to `BottomSheetProps` in `BottomSheet.tsx` with default `true`.
>   - **Functions**:
>     - Updates `onChange` in `BottomSheet` to handle index conversion based on `enableDynamicSizing`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=draftbit%2Freact-native-jigsaw&utm_source=github&utm_medium=referral)<sup> for aa95f6be021c2d0030005a8d2f7a1a0f0c3a706e. It will automatically update as commits are pushed.</sup>


packages/core/src/components/BottomSheet/BottomSheet.tsx
```diff
@@ -109,7 +109,7 @@ const BottomSheet = React.forwardRef<BottomSheetComponent, BottomSheetProps>(
           borderWidth,
           borderColor: borderColor ?? theme.colors.border.brand,
         }}
-        onChange={(index) => onSettle?.(mappedSnapPoints.length - index - 1)}
+        onChange={(index) => onSettle?.(index)}
```
YoussefHenna, line 112: I don't think this is the issue. Looking at the linear issue (and testing on my own), seems the bottom sheet has 4 snap points, even though by default there should only be 3 (top, middle, bottom). I think this inconsistency is what is leading to the incorrect index, but the existing logic for this is still correct.

YoussefHenna: Seems there's a default bottommost snap point for some reason

YoussefHenna: I can investigate this later today, but let me know if you have any idea why this could be happening.

sieu-db: Ya, I will investigate more today 🙏 thanks for point out

YoussefHenna (approved):
Nice find!

PR #964, opened 2024-10-31
Proper typography support in themes

YoussefHenna:
## Main changes
- Adds proper typing for theme's typography
- Updates theme object proxy logic to also work with typography (previously only worked with colors)
- Updated default typography values to match defaults we use on the platform codegen
  - Will have no noticeable effect, the defaults in codegen already always overrode the default values in jigsaw.
  
## Other changes
- Theme object validation became more complex with the addition of typography. Removed manual type checks and replaced with `zod` as was suggested by @needs when initially working on colors for themes.
- Use `deepmerge-ts` instead of `deepmerge` since `deepmerge-ts` allows for better customization on how merging is done, which is was needed here. Also `deepmerge` seems to be abandoned while this one is actively maintained.
- Added a simple example of varying typography with themes.

----

> [!IMPORTANT]
> Adds typography support to themes, updates default values, and simplifies validation using `zod`.
> 
>   - **Behavior**:
>     - Adds typography support to theme object proxy logic in `Provider.tsx` and `createThemeValuesProxy.ts`.
>     - Updates default typography values in `DefaultTheme.ts` to match platform codegen defaults.
>   - **Validation**:
>     - Replaces manual type checks with `zod` in `validators.ts` for theme validation.
>   - **Dependencies**:
>     - Replaces `deepmerge` with `deepmerge-ts` in `createTheme.ts` for better merge customization.
>   - **Examples**:
>     - Adds typography example in `ThemeExample.tsx`.
>   - **Tests**:
>     - Updates tests in `createThemeValuesProxy.test.ts` and `validators.test.ts` to cover typography.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=draftbit%2Freact-native-jigsaw&utm_source=github&utm_medium=referral)<sup> for 662c0a5b49a208c5203603f64ca6e49505cc13a3. It will automatically update as commits are pushed.</sup>


package.json
```diff
@@ -134,8 +134,5 @@
       "not samsung <= 4"
     ],
     "node": "16"
-  },
-  "engines": {
```
sieu-db, line 138: Curious why we need remove this

packages/theme/src/__tests__/createThemeValuesProxy.test.ts
```diff
@@ -315,4 +337,50 @@ describe("createThemeValuesProxy tests", () => {
       );
     });
   });
+
+  describe("Typography Values", () => {
+    test("returns typography object when value is top level", () => {
+      const proxied = createThemeValuesProxy(
```
sieu-db, line 343: Recommend change this createThemeValuesProxy prop support params as object, not by multiple params with index like this

Reason is it will really hard to understand/update which prop doing what when look at this, also it's easy when enhance/scale this function, because we don't need to remember index of each params

Example by below

```suggestion
      const proxied = createThemeValuesProxy(
        value: value,
        style: {},
        fontWeight: 400,
        platform: "android",
        themeColor: "light"
      )
```

package.json
YoussefHenna: We don't 'need' to, but I figured no reason to limit it to node 18. Have to keep switching versions back and forth locally. No reason it won't work with higher node versions

packages/theme/src/__tests__/createThemeValuesProxy.test.ts
YoussefHenna: Makes sense, I'll do that change

YoussefHenna: Done!

sieu-db (dismissed):
LGTM 💯

YoussefHenna:
Keeping un-merged until I implement changes into the builder. In case I discover something else that needs updating.

PR #967, opened 2024-11-15
Set default width for linear progress component

sieu-db:
## Overview

Simulator Screenshot - iPhone 15 - 2024-11-15 at 12 46 01


YoussefHenna (approved):
What was the behavior before? Did it just show nothing until you set a width? We should also update the mapping file to show this as a default style.

Looks good in any case!

sieu-db:
> What was the behavior before? Did it just show nothing until you set a width? We should also update the mapping file to show this as a default style.
> 
> Looks good in any case!

Yes, correct, it's show nothing

PR #968, opened 2024-11-18
Support expo image

sieu-db:
## Overview
- Support expo-image
- Provide the method convert from image to expoImage without any migrations
- Support some additional props for expo-image (transition, blurhash ...etc...)
- Add expo image example
- Support remote SVG & local SVG
- Support correctly for web also

## Demo

https://github.com/user-attachments/assets/9b3d8aad-1c3b-4614-87f5-ff81f3a6bfd2

----

> [!IMPORTANT]
> Add support for `expo-image` with new features and examples, updating `Image.tsx` and adding `ExpoImageExample.tsx`.
> 
>   - **Behavior**:
>     - Replace `NativeImage` with `ExpoImage` in `Image.tsx` to support `expo-image` features.
>     - Add support for `transition`, `blurhash`, `contentFit`, `contentPosition`, `cachePolicy`, `allowDownscaling`, and `recyclingKey` in `Image.tsx`.
>     - Add `ExpoImageExample.tsx` to demonstrate `expo-image` features, including remote/local images, aspect ratios, and SVGs.
>   - **Dependencies**:
>     - Add `expo-image` version `1.10.6` to `package.json`.
>   - **Misc**:
>     - Add `ExpoImageExample` to `App.tsx` routes.
>     - Export `ExpoImage` in `index.tsx` of `core` and `ui` packages.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=draftbit%2Freact-native-jigsaw&utm_source=github&utm_medium=referral)<sup> for 1397b7aea60a0c0b7837cc89bf85ee99a9b3ae7e. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/packages/core/src/components/ExpoImage.tsx
+++ b/packages/core/src/components/ExpoImage.tsx
@@ -16,20 +16,16 @@ type ImageStyleProp = {
 };
 
 interface ExtendedImageProps extends ExpoImageProps {
-  transition?:
-    | number
-    | {
-        duration?: number;
-        effect?:
-          | "cross-dissolve"
-          | "flip-from-top"
-          | "flip-from-right"
-          | "flip-from-bottom"
-          | "flip-from-left"
-          | "curl-up"
-          | "curl-down";
-        timing?: "ease-in-out" | "ease-in" | "ease-out" | "linear";
-      };
+  transitionDuration?: number;
+  transitionEffect?:
+    | "cross-dissolve"
+    | "flip-from-top"
+    | "flip-from-right"
+    | "flip-from-bottom"
+    | "flip-from-left"
+    | "curl-up"
+    | "curl-down";
+  transitionTiming?: "ease-in-out" | "ease-in" | "ease-out" | "linear";
   contentFit?: "cover" | "contain" | "fill" | "none" | "scale-down";
   contentPosition?: ImageContentPosition;
   cachePolicy?: "none" | "disk" | "memory" | "memory-disk";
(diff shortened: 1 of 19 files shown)


packages/core/src/components/Image.tsx
```diff
@@ -68,13 +114,22 @@ const Image: React.FC<ImageProps> = ({
     styles as ImageStyleProp
   );
 
+  // Use contentFit if provided, otherwise fall back to resizeMode
+  const finalContentFit = contentFit || resizeModeToContentFit(resizeMode);
```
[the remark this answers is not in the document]
sieu-db: Make sense 👍

packages/core/src/components/ExpoImage.tsx
```diff
@@ -0,0 +1,165 @@
...
+  cachePolicy = "memory-disk",
+  allowDownscaling = true,
+  blurRadius,
+  blurhash,
+  ...props
+}) => {
+  let imageSource =
+    source === null || source === undefined
+      ? Config.placeholderImageURL
+      : source;
```
YoussefHenna, line 103:
```suggestion
  const imageSource = source ?? Config.placeholderImageURL
```

packages/core/src/components/ExpoImage.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  const transition = {
+    timing: transitionTiming,
+    duration: transitionDuration,
+    effect: transitionEffect,
+  };
+
+  if (aspectRatio) {
+    return (
+      <AspectRatio style={[style, { width, height, aspectRatio }]}>
```
YoussefHenna, line 122: `style` is applied here, but then again in the `Image` component. This might cause issues where the same style is applied twice. Think padding or margin for example, it'll be applied twice.

I think it might be better to just render the `Image` as is, and instead of wrapping in an aspect ratio component, maybe we can copy over the aspect ratio logic here, or expose it as a common function that both can use?

YoussefHenna (commented):
Looks great overall, but I think we need to address the double style thing before merging.

packages/core/src/components/ExpoImage.tsx
YoussefHenna: Simpler?

